PR #82, opened 2014-04-29
DATACMNS-497 Changed visibility to protected for customization

dwelch2344:
Just filed [DATACMNS-497](https://jira.spring.io/browse/DATACMNS-497) for this minor change. 

I've got a scenario where I can't create repositories without overriding `RepositoryFactorySupport.getRepositoryMetadata()` and can't do so because it has default visibility.

Also, I've signed the Contributor Agreement so I should be good there. Thanks!


olivergierke:
This seems to have been fixed with the patch for [DATACMNS-631](https://jira.spring.io/browse/DATACMNS-631).

PR #110, opened 2015-01-19
DATACMNS-634 - Scan for repositories of superclasses.

thomaseizinger:
I believe the referenced issue is wrongly tagged as an issue in SDR as the affected code belongs to the spring-data-commons repository. Previously, a repository defined for a supertype of a class could not be discovered. This pull request adds the functionality to retrieve the repository information of the repository assigned to the supertype.


olivergierke:
Good catch. I've moved the ticket to Spring Data Commons and applied the PR with minor polishing.

PR #111, opened 2015-01-20
DATACMNS-621 - QSort treats long paths incorrectly.

christophstrobl:
We now inspect intermediate path elements to build sort order accordingly.

### DIFF
--- a/src/test/java/org/springframework/data/querydsl/WrapperToWrapWrapperForUserWrapper.java
+++ b/src/test/java/org/springframework/data/querydsl/WrapperToWrapWrapperForUserWrapper.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


src/test/java/org/springframework/data/querydsl/WrapperToWrapWrapperForUserWrapper.java
```diff
@@ -0,0 +1,29 @@
...
+import com.mysema.query.annotations.QueryEntity;
+import com.mysema.query.annotations.QueryInit;
+
+/**
+ * @author Christoph Strobl
+ */
+@QueryEntity
+public class WrapperToWrapWrapperForUserWrapper {
+
+	@QueryInit("wrapper.user")//
```
wimdeblauwe: 2 slashes are probably not needed?

christophstrobl: yes and no, spring data code formatter would otherwise move everything into one line

PR #116, opened 2015-02-24
DATACMNS-650 - Provide a CloseableIterator abstraction as a foundation for streaming of results.

thomasdarimont:
Introduced CloseableIterator, AutoCloseableIterator and ForwardingAutoCloseableIterator to bridge between the two.

### DIFF
--- a/src/main/java/org/springframework/data/util/ReflectionUtils.java
+++ b/src/main/java/org/springframework/data/util/ReflectionUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2013 the original author or authors.
+ * Copyright 2012-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.util.stream.Stream;
 
 import org.springframework.beans.BeanUtils;
 import org.springframework.core.annotation.AnnotationUtils;
@@ -30,9 +31,25 @@ import org.springframework.util.ReflectionUtils.FieldFilter;
  * Spring Data specific reflection utility methods and classes.
  * 
  * @author Oliver Gierke
+ * @author Thomas Darimont
  * @since 1.5
  */
 public abstract class ReflectionUtils {
+
+	private static final Class<?> JAVA8_STREAM_TYPE;
+
+	static {
+		String streamClassName = "java.util.stream.Stream";
+
+		Class<?> cls = null;
+		if (ClassUtils.isPresent(streamClassName, ReflectionUtils.class.getClassLoader())) {
+			try {
+				cls = Class.forName(streamClassName, true, ReflectionUtils.class.getClassLoader());
+			} catch (ClassNotFoundException ignore) {}
+		}
+
+		JAVA8_STREAM_TYPE = cls;
+	}
 
 	private ReflectionUtils() {}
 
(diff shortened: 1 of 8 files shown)


jhoeller:
Guys, java.io.Closeable extends java.lang.AutoCloseable anyway... So I'm not sure what AutoCloseableIterator really adds here, and why the forwarding is needed?

We're implementing Java 6's Closeable to implicitly get Java 7 AutoCloseable behavior in a few places in the core framework as well. Seems to work fine for those purposes.

thomasdarimont:
`AutoCloseable` was just introduced with Java 7 so it isn't available if SD Commons is used with Java 6.

jhoeller:
Sure, but my point is: Any of our classes implementing Closeable implicitly implements AutoCloseable on Java 7, since the Closeable interface itself happens to extend AutoCloseable on Java 7+. This is perfectly Java 6 compatible and also enables AutoCloseable when running on Java 7, simply through implementing Closeable. Any extra business with AutoCloseable decoration is effectively unnecessary.

thomasdarimont:
A gotcha :) You are right.

src/main/java/org/springframework/data/util/ReflectionUtils.java
```diff
@@ -30,10 +31,26 @@
...
  */
 public abstract class ReflectionUtils {
 
+	private static final Class<?> JAVA8_STREAM_TYPE;
+
+	static {
+		String streamClassName = "java.util.stream.Stream";
+
+		Class<?> cls = null;
+		if (ClassUtils.isPresent(streamClassName, ReflectionUtils.class.getClassLoader())) {
```
olivergierke: Can't we just use `ClassUtils.forName(…)` here?

olivergierke:
As outline in the comments on the [original pull request](https://github.com/spring-projects/spring-data-mongodb/pull/274#issuecomment-75436060) I think we should focus on a `CloseableIterable` as return type. Using an `Iterator` as return type seems rather uncommon.

thomasdarimont:
After having played a bit with the API I still think that a `CloseableIterable<T>`provides a richer API experience.

Consider the following variants:
```java
//Variant 1 - CloseableIterator<Person>
@Query("{firstname:{$in:?0}}")
CloseableIterator<Person> findByCustomQueryByFirstnames(List<String> firstnames);

//V1: Use case 1 - loop over all results.
try (CloseableIterator<Person> result = repository.findByCustomQueryByFirstnames(Arrays.asList("Dave"))){
	for (Person person : result) {
		readPersons.add(person);
	}
}

//V1: Use case 2 - check for results and process later
try (CloseableIterator<Person> result = repository.findByCustomQueryByFirstnames(Arrays.asList("Dave"))){
	if(result.hasNext()){
		//doSomething
	}else{
		//doSomething else
	}

	for (Person person : result) {
		readPersons.add(person);
	}	
}


//Variant 2 - CloseableIterable<Person>
@Query("{firstname:{$in:?0}}")
CloseableIterable<Person> findByCustomQueryByFirstnames(List<String> firstnames);

//V2: Use case 1 - loop over all results.
try (CloseableIterable<Person> result = repository.findByCustomQueryByFirstnames(Arrays.asList("Dave"))){
	for (Person person : result) {
		readPersons.add(person);
	}
}

//V2: Use case 2 - check for results and process later
try (CloseableIterable<Person> result = repository.findByCustomQueryByFirstnames(Arrays.asList("Dave"))){

    Iterator iter = result.iterator();

	if(iter.hasNext()){
		//doSomething
	} else{
		//doSomething else
	}

	for (Person person : result) {
		readPersons.add(person);
	}	

	//what to close now? iterator or result?
}
````
In Variant1 it is clear what the "resource" and a user can conveniently iterate over the results.
Variant2 however doesn't make this clear per se - IMHO.

PR #122, opened 2015-05-13
Change property reflection mechanisms from java.beans.Introspector to spring specific BeanUtils

tawek:
The purpose is to:
 1. cache the introspection results
 2. enable BeanInfoFactory support to custom bean reflection strategies (fluent setters, default methods with getters/setters etc.)

DATACMNS-693


olivergierke:
That's merged, thanks!

PR #129, opened 2015-06-17
DATACMNS-712 - Support CompletableFuture as Repository return type.

thomasdarimont:
Added QueryExecutionConverter for CompletableFuture.
This should work out of the box if used in combination with Spring 4.2.0 (RC2 / BUILD-SNAPSHOT).


thomasdarimont:
I saw that this is already done - strange :)

PR #135, opened 2015-07-27
DATACMNS-739 - Performance improvement for MultiTransactionStatus

ameyjadiye:
Its always better in Multi-threading and low latency application that not to have single lock on Map.
ConcurrentHashMap holds default 16 locks which significantly improves the performance than having single lock SyncronizedMap.

### DIFF
--- a/src/main/java/org/springframework/data/transaction/MultiTransactionStatus.java
+++ b/src/main/java/org/springframework/data/transaction/MultiTransactionStatus.java
@@ -37,7 +37,8 @@ import org.springframework.util.Assert;
 class MultiTransactionStatus implements TransactionStatus {
 
 	private final PlatformTransactionManager mainTransactionManager;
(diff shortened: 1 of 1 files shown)


olivergierke:
Do you have any before / after comparisons? `MultiTransactionStatus` hasn't showed up in profilings so far.

@jexp - Opinions?

jexp:
I think a perf test would make sense. 

You only benefit from the 16 segments if you have that high concurrency level.

ameyjadiye:
ok, let me put the perf results here .. also let me make it configurable so programmers can opt for number of locks, if configured as 1 its nothing but the simple  _HashMap_, and default should be _concurrencyLevel(1)_ , if some one using it in multi-threading environment programmers can increase it.

olivergierke:
Is there any further feedback? If not I'd stick to the current implementation and close this one without merge.

ameyjadiye:
Hi @olivergierke , I totally forgot to work on this and this was sitting in my **todo** from long , let me pull some comparison stats and junits if possible - thanks for not closing this.

mp911de:
Closing this issue after a significant period of inactivity.

PR #138, opened 2015-08-14
[Test case DATACMNS-753] parsing of Sort fails to fallback properly

Shredder121:
The test cases of `DATACMNS-408` included a input of `null`, but a request with `?sort=` means that the parameter is `""`.
Discussion https://jira.spring.io/browse/DATACMNS-753.


olivergierke:
That's merged and a fix applied, thanks!

PR #140, opened 2015-08-17
s/User/Person/

martin-g:
It looks odd (wrong) that `PersonRepository` deals with `User` entity.
I think the type parameter should really be `Person`


olivergierke:
Good catch, thanks.

martin-g:
Is it good? :)
Because you didn't merge it but just closed it.

martin-g:
Oh. NVM.

PR #142, opened 2015-08-26
Make class public

yakuninv:
It has package visibility which prevents this adapter from using outside the package it declared in.


olivergierke:
Closing due to lack of feedback.

PR #143, opened 2015-08-28
changed return type of findAll methods in QueryDslPredicateExecutor to List

lpandzic:


olivergierke:
What's the reasoning behind this change? We can't simply change return types as this will inevitably break implementations. Also `Iterable` was used by design to allow for implementations that are not `List` based in the first place.

lpandzic:
Main issue with `Iterable` is it's API which is really lacking compared to `Collection` or `List` API.

For instance, to convert `List` to a `Stream` all you have to do is call `#stream` available from `Collection` interface. On the other hand, converting `Iterable` to a `Stream` is not so simple: `StreamSupport.stream(iterable.spliterator(), false)`.

Also, API of Repositories is not consistent when we take `JPARepository` into account. 
As an alternative to this PR, would it be more acceptable to provide a specialized JPAQueryDslPredicateExecutor which overrides those methods with specialized return type? `QueryDslJpaRepository` already has List as it's return type for those methods.

pivotal-issuemaster:
@lpandzic Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=143) to manually synchronize the status of this Pull Request.

lpandzic:
Since https://jira.spring.io/browse/DATACMNS-867 will solve the issue of converting to Stream, I'm closing this pull request.

PR #145, opened 2015-10-05
Typo

ricardoespsanto:
Targte instead of Target


olivergierke:
Thanks, that's merged! :)

PR #148, opened 2015-11-18
ChainedTransactionManager integration tests

mit-jones:
There looks to be an issue with the ChainedTransactionManager.

Two integration tests (ChainedTransactionManagerIntegrationTests) have been added both functionally the same, both should pass. However one fails (with a CannotCreateTransactionException) where it calls a method that has been advised with transaction propagation NOT_SUPPORTED.

I think there is an issue with the ChainedTransactionManager possibly in the getTransaction() method where the synchronizationManager is initialised regardless of the transaction propagation. Compare this with AbstractPlatformTransactionManager.getTransaction() where initSynchronization() would only be called for transaction propagation of REQUIRED, REQUIRES_NEW and NESTED.

This issue is probably related to https://jira.spring.io/browse/DATACMNS-370 so will be added to this existing JIRA

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,22 +94,24 @@
 			<optional>true</optional>
 		</dependency>
 
+		<!--  Querydsl -->
+
 		<dependency>
-			<groupId>com.mysema.querydsl</groupId>
+			<groupId>com.querydsl</groupId>
 			<artifactId>querydsl-core</artifactId>
 			<version>${querydsl}</version>
 			<optional>true</optional>
 		</dependency>
 
 		<dependency>
-			<groupId>com.mysema.querydsl</groupId>
+			<groupId>com.querydsl</groupId>
 			<artifactId>querydsl-apt</artifactId>
 			<version>${querydsl}</version>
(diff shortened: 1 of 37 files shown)


olivergierke:
I don't think we can merge this as we're not gonna introduce dependencies to JDBC related artifacts in Spring Data Commons. Still would like to get @jexp's opinion on that one.

mit-jones:
Okay understood but how do we proceed from here, I believe the issue still remains. I do not know another way of providing a reproducible test case that can highlight this issue. If you can give me some guidance I am okay with trying another approach but I think it will be difficult to prove without an integration test.

pivotal-issuemaster:
@mit-jones Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=148) to manually synchronize the status of this Pull Request.

schauder:
As @olivergierke explained above we won't merge this one. This does not affect the underlying issue.

PR #151, opened 2016-01-14
DATACMNS-658 - Add 'ignorecase' ordering option to SortHandlerMethodArgumentResolver

dan-nawrocki:
I've found the need to be able to turn on ```Order#ignoreCase``` when using the sorting parameters from Spring Data REST. This pull request recognizes ```,ignorecase``` when added to the end of a web ```sort``` parameter to turn on this feature. For some examples:

```http://localhost/repository?sort=prop,ignorecase```: Returns all entries in the ```repository``` sorted (asc) by ```prop```, ignoring case.

```http://localhost/repository?sort=prop,desc,ignorecase```: Returns all entries in the ```repository``` sorted descending by ```prop```, ignoring case.

This pull request is backwards compatible when ```,ignorecase``` is not used.

As I mentioned above, I need this in support of Spring Data REST. I don't see why it would break anything else, but I don't have a very deep understanding of the entire Spring Data ecosystem to know for sure. If there are any problems with this pull request or my approach in general I'd be happy to fix them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>1.11.2.RELEASE</version>
+	<version>1.11.3.BUILD-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
 	<parent>
 		<groupId>org.springframework.data.build</groupId>
 		<artifactId>spring-data-parent</artifactId>
-		<version>1.7.2.BUILD-SNAPSHOT</version>
+		<version>1.7.2.RELEASE</version>
 	</parent>
 
 	<properties>
@@ -257,8 +257,8 @@
 
 	<repositories>
 		<repository>
-			<id>spring-libs-snapshot</id>
-			<url>https://repo.spring.io/libs-snapshot</url>
+			<id>spring-libs-release</id>
+			<url>https://repo.spring.io/libs-release</url>
 		</repository>
 	</repositories>
 
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>1.11.1.RELEASE</version>
+	<version>1.11.2.BUILD-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>1.11.0.RELEASE</version>
+	<version>1.11.1.BUILD-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
--- a/pom.xml
+++ b/pom.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 
 	<modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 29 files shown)


olivergierke:
While I appreciate the effort, I don't think this is a solution that we can use. So far, clients have been used  to the fact that the last segment is considered the order in case its either `asc` or `desc`. So if a client previously sent `foo,desc,ignorecase` because all `foo`, `desc` and  `ignorecase` were *properties* of the domain, that request would now get a different semantic without the client noticing.

I guess we're gonna need a new format entirely that can be clearly differentiated from the current one that can get the additional option added. See further discussion at the related [ticket](https://jira.spring.io/browse/DATACMNS-658).

dan-nawrocki:
Thanks for reviewing the PR. I see what you mean with respects to the situation where all 3 are domain properties.

From one point of view, even treating last parameter as a sort order if it is either ```asc``` or ```desc``` falls short of the desired goal as well. If a client sends ```foo,desc``` and expects sorting by two properties, namely ```foo``` and ```desc```, he would be surprised to find that the sorting was actually performed only by ```foo``` (and then probably in the opposite direction as expected). The documentation is clear on this behavior, but the point still remains -- you can't sort by a property named ```asc``` or ```desc``` unless you add additional sorting terms at the end.

I see your point about needing a new format and agree wholeheartedly. Here's one proposal to borrow from Spring Data REST:
* each sort parameter is in the format ```property,direction,flag1,flag2,...,flagN```
  * ```property``` is the name of the property and is mandatory
  * ```direction``` is the sort direction (either ```asc``` or ```desc```) and is optional
  * ```flag1,flag2,...,flagN``` is an optional keyword to specify various options. If the client wants to provide one or more flags, then the ```direction``` *must* be specified. Supported flags could be:
    * ```ignorecase``` to turn on case-insensitive sorting
    * ```nullfirst``` to turn on null-first sorting
    * ```nulllast``` to turn on null-last sorting
* if more than one sort direction is needed, multiple ```sort``` URL parameters should be used
* if the provided sort parameter does not match this format, an exception would be thrown. The situations are:
  * 0 values provided after splitting at the ```,``` delimiter
  * ```direction```, if provided, is neither ```asc``` or ```desc```
  * any of the ```flag```s are not a recognized flag value
  * conflicting behavior between two ```flag```s (e.g,. specifying ```nullfirst``` and ```nulllast```)
* this has the advantage of being able to sort properties in different directions (e.g., ```foo``` sorted ```asc``` and ```bar``` sorted ```desc```)

This would obviously break existing clients, so a new URL parameter is needed in order to maintain backwards compatibility. ```sort2``` perhaps? An upgrade path for users would be something like:

```?sort=prop1``` would become ```?sort2=prop1```

```?sort=prop1,desc``` would become ```?sort2=prop1,desc```

```?sort=prop1,prop2``` would become ```?sort2=prop1&sort2=prop2```

```?sort=prop1,prop2,desc``` would become ```?sort2=prop1,desc&sort2=prop2,desc```

New features supported by this would be:

```?sort2=prop1,asc&sort2=prop2,desc``` would sort ```asc``` by ```prop1```, then ```desc``` by ```prop2```

```?sort2=prop1,asc,ignorecase``` would sort ```asc``` by ```prop1``` using case-insensitive sorting

```?sort2=prop1,asc,nullfirst``` would sort ```asc``` by ```prop1``` using null-first sorting

Thoughts?

dan-nawrocki:
@olivergierke any comments? I've got some slack time at work coming up and would love to get this implemented.

wfaithfull:
I'll just add my voice to this request.

I whole-heartedly agree with the need to enforce backwards compatibility. In lieu of changing the semantics of the `sort` parameter,  could we not simply also support a parameter called, for example, `sortIgnoreCase`? 

Then we could just do `?sortIgnoreCase=foo,desc`, and existing requests to `?sort=foo,desc` would be untouched.

dan-nawrocki:
@wfaithfull Yes, backwards compatibility is extremely important. The intent of the ```sort2``` parameter as described above does exactly what you propose by (a) leaving the original ```sort``` as-is, and (b) only adding new functionality to ```sort2```. I've proposed adding ```nullfirst``` and ```nulllast``` in addition to ```ignorecase``` because it's a good semantic fit and someone may find it useful sometime.

@olivergierke has there been any more consideration of this? I'm happy to submit pull requests if there's a chance this would be merged.

wfaithfull:
@dan-nawrocki Ah, I see. You want to call it `sort2` to leave it open to future extension.

Well, the scheme I currently follow in my application is a custom [SortHandlerMethodArgumentResolver](https://gist.github.com/anonymous/59eac9441b6860c3daea0cfc4827378e) which looks for an extra parameter in the request called `ignoreCase`, for example, `?sort=foo,desc&ignoreCase`, which is fine, but I'd rather we had a consistent approach.

dan-nawrocki:
Closing in favor of adding this to the 1.12.x branch.

And because I messed up the base for this PR :(

PR #152, opened 2016-01-20
DATACMNS-804 - Fix ParametersParameterAccessor's iterator with non-bindables.

simonbasle:


simonbasle:
fixed by @olivergierke in 85e569b

PR #153, opened 2016-01-28
DATACMNS-810 - Add core types for Query By Example support.

christophstrobl:
Added required types for Query By Example which should be used by the individual store implementations. 

Generally the `Example` captures a sample object and allow various settings concerning the mapping into an actual query. So there’s configuration options for handling null values, string matching, property paths to ignore,...

----

Required by: DATAJPA-218, DATAMONGO-1245


src/main/java/org/springframework/data/domain/Example.java
```diff
@@ -0,0 +1,113 @@
...
+	private final T probe;
+	private final ExampleSpec<? extends T> exampleSpec;
+
+	/**
+	 * Create a new {@link Example} including all non-null properties by default.
+	 *
+	 * @param probe The probe to use. Must not be {@literal null}.
+	 */
+	@SuppressWarnings("unchecked")
+	public Example(T probe) {
```
olivergierke: Make private to make sure `of(…)` is the only way to create instances of `Example`.

src/main/java/org/springframework/data/domain/Example.java
```diff
@@ -0,0 +1,113 @@
...
+		this.exampleSpec = ExampleSpec.of((Class<T>) probe.getClass());
+	}
+
+	/**
+	 * Create a new {@link Example} including all non-null properties by default.
+	 *
+	 * @param probe The probe to use. Must not be {@literal null}.
+	 * @param exampleSpec The example specification to use. Must not be {@literal null}.
+	 */
+	public Example(T probe, ExampleSpec<? extends T> exampleSpec) {
```
olivergierke: Make private to make sure `of(…)` is the only way to create instances of `Example`.

src/main/java/org/springframework/data/domain/Example.java
```diff
@@ -0,0 +1,113 @@
...
+		return (Class<T>) ClassUtils.getUserClass(probe.getClass());
+	}
+
+	/**
+	 * Create a new {@link Example} including all non-null properties by default.
+	 *
+	 * @param probe must not be {@literal null}.
+	 * @return
+	 */
+	public static <T> Example<T> of(T probe) {
```
olivergierke: Move factory methods up right underneath the constructors.

src/main/java/org/springframework/data/domain/ExampleSpec.java
```diff
@@ -0,0 +1,797 @@
...
+	 * @param genericPropertyMatcher callback to configure a {@link GenericPropertyMatcher}, must not be {@literal null}.
+	 * @return
+	 */
+	public ExampleSpec<T> withMatcher(String propertyPath, GenericPropertyMatcher genericPropertyMatcher) {
+
+		Assert.hasText(propertyPath, "PropertyPath must not be empty!");
+		Assert.notNull(genericPropertyMatcher, "GenericPropertyMatcher must not be empty!");
+
+		PropertySpecifiers propertySpecifiers = new PropertySpecifiers(this.propertySpecifiers);
+		PropertySpecifier propertySpecifier = new PropertySpecifier(propertyPath);
```
olivergierke: Make `PropertySpecifier` an immutable type.

src/main/java/org/springframework/data/domain/ExampleSpec.java
```diff
@@ -0,0 +1,797 @@
...
+				defaultIgnoreCase);
+	}
+
+	/**
+	 * Create a new {@link Example} containing the {@code probe}.
+	 *
+	 * @param probe must not be {@literal null}.
+	 * @return
+	 */
+	public Example<T> createExample(T probe) {
```
olivergierke: Can we avoid this method being exposed or maybe even remove it at all? We're creating ambiguity here which method should be preferred to be used (`Example.of(…)` or this one here).

src/main/java/org/springframework/data/domain/ExampleSpec.java
```diff
@@ -0,0 +1,797 @@
...
+		private StringMatcher(Type type) {
+			this.type = type;
+		}
+
+		/**
+		 * Get the according {@link Part.Type}.
+		 *
+		 * @return {@literal null} for {@link StringMatcher#DEFAULT}.
+		 */
+		public Type getPartType() {
```
olivergierke: We can't refer to the `query.parser` package here as it's a less fundamental package and would introduce a cycle. I couldn't really find usage of the method in Spring Data Commons itself. Maybe we can invert the dependency by introducing a translating mechanism in the `query.parser` package?

src/main/java/org/springframework/data/domain/ExampleSpecAccessor.java
```diff
@@ -0,0 +1,149 @@
...
+
+import java.util.Collection;
+
+/**
+ * Accessor for the {@link ExampleSpec} to use in modules that support query by example (QBE) querying.
+ *
+ * @author Mark Paluch
+ * @since 1.12
+ */
+public class ExampleSpecAccessor {
```
olivergierke: What exactly is the purpose of this class? Looks like it's publicly exporting some methods that are package private on ExampleSpec? It doesn't really fit into the domain package as it's designed to only contain types to be used by users, not SPIs in general.

src/main/java/org/springframework/data/repository/query/QueryByExampleExecutor.java
```diff
@@ -0,0 +1,85 @@
...
+public interface QueryByExampleExecutor<T> {
+
+	/**
+	 * Returns a single entity matching the given {@link Example} or {@literal null} if none was found.
+	 *
+	 * @param Example can be {@literal null}.
+	 * @return a single entity matching the given {@link Example} or {@literal null} if none was found.
+	 * @throws org.springframework.dao.IncorrectResultSizeDataAccessException if the Example yields more than one result.
+	 */
+	<S extends T> T findOne(Example<S> example);
```
olivergierke: I guess this method has to return an instance of `S` to stay consistent with `CrudRepository.findOne(…)`? Same for all other methods?

src/main/java/org/springframework/data/domain/ExampleSpecAccessor.java
mp911de: This class is used by the individual modules to create queries and to access details stored in Example/ExampleSpec. How would have solved it? Put it to query.spi and add public getters to ExampleSpec?

olivergierke: I guess the question is whether we need it at all if moving the class out of this package would require us to make certain methods in `ExampleSpec` public anyway.

mp911de: This class comes handy in the modules as it contains code that would be duplicated in the modules otherwise. We should be able to get along by putting this class in a different package and making certain methods public in `ExampleSpec`.

src/main/java/org/springframework/data/repository/query/QueryByExampleExecutor.java
mp911de: Had a second look at the declaration. In fact, a Repository will always return `T` because the type is looked up from the repository, not the probe. I'd change the declaration to `T findOne(Example<? extends T> example)`, same for the other methods.

olivergierke: Why should we drop that additional boundary? Assume I have `Contact` and a `Person` extending the former. If I hand in an `Example<Person>`, why shouldn't I get a `Person` returned but a `Contact`?

mp911de: Not sure whether I get you right. IIRC `Repository<Contact>` works that way that the resulting type is set from the repository definition when using methods declared on the provided repository interfaces (i. e. `MongoRepository`). 

When using `<S extends T> S findOne(Example<S>)` the query by example methods would use the `S` entity type and behave like `<S extends T> List<S> save(Iterable<S> entites)`?

src/main/java/org/springframework/data/domain/Example.java
mp911de: Done

src/main/java/org/springframework/data/domain/ExampleSpec.java
mp911de: Done.

mp911de:
Refactored Query by Example API to support typed and untyped ExampleSpec's. 

Most significant changes for Spring Data JPA:
https://github.com/spring-projects/spring-data-jpa/pull/164/files#diff-b23bd4fdc1c19109c8f6830041fdc651R437

Test showing the usage related to inheritance:
https://github.com/spring-projects/spring-data-jpa/pull/164/files#diff-90a1e1093407b10052e14478211df4d4R1290

PR #155, opened 2016-02-20
DATACMNS-820 - Allow setter invocations in projections backed by beans.

mp911de:
Invocations to setter methods are passed now to the backing bean.


gregturn:
I thought projections were read only.

mp911de:
@gregturn Projections can already be written if backed by a map

PR #157, opened 2016-03-23
DATACMNS-831 - AbstractMappingContext.getPersistentEntity sometimes returns an empty entity when called by multiple threads

mklamra:
There is a race condition in AbstractMappingContext that causes the issue described in DATACMNS-831. This pull request contains a fix for the bug.


olivergierke:
This has been polished and merged, thanks!

PR #159, opened 2016-04-14
DATACMNS-809 — Investigate use of method handles for property access

mp911de:
We now support generated PersistentPropertyAccessors when using Java 7 and if property/association name hashCodes are unique within a PersistentEntity.
Generated PersistentPropertyAccessors provide optimized access to properties. They use either MethodHandles or direct property/field access, depending on the visibility/final modifiers of the entity type and its members. A generated PersistentPropertyAccessor is injected into the originating class loader of the entity class to enable optimizations for package-default/protected member access.

## Open issues

- [x] Where to keep the root reference to `ClassGeneratingPropertyAccessorFactory` as creating factories per `PersistentEntity` might be expensive

----
Related ticket: [DATACMNS-809](https://jira.spring.io/browse/DATACMNS-809)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>1.13.0.BUILD-SNAPSHOT</version>
+	<version>1.13.0.DATACMNS-809-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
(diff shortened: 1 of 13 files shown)


olivergierke:
I've applied a round of polishing. Open todo as just discussed: move registration of `PersistentPropertyAccessorFactory` into `AbstractMappingContext` to avoid creating instances for every `BasicPersistentEntity`.

I tested the current state of implementation against the Spring Data examples and they seem to work fine also in combined usage with the Spring Boot development tools. I noticed that `ClassGeneratingPropertyAccessorFactory.canGenerateAccessorClass(…)` was called quite often (once per request for a `PersistentPropertyAccessor` but I guess that's gonna resolve with the refactoring suggested above, as the call to that method will be a one time effort at the time the `PersistentEntity` is created.

PR #160, opened 2016-05-06
DATACMNS-809 - Reproduce issue accessing persistent properties from interface-based entities using the ClassGeneratingPropertyAccessorFactory.

jxblum:
Test class demonstrating issue.


olivergierke:
That should be fixed in the mean time!

PR #161, opened 2016-05-06
DATACMNS-853 - Support interface entity types using generated property accessors.

mp911de:
Property accessors (getter and setter) defined on an interface require a different instruction opcode for invoking methods. We now distinguish whether a property accessor is defined on an interface and use either INVOKEINTERFACE or INVOKEVIRTUAL otherwise.

----

Related tickets: [DATACMNS-853](https://jira.spring.io/browse/DATACMNS-853), [SGF-494](https://jira.spring.io/browse/SGF-494)
Related pull request: #160

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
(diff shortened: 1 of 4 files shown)


olivergierke:
That's merged, thanks!

PR #162, opened 2016-05-17
DATACMNS-854 - Allow custom implementation of Repository methods with…

terje2001:
Add additional check for finding customized method where method contains generics. Overriding methods without generics already works, this patch simply makes the behavior consistent.


olivergierke:
That's been merged a while ago :).

PR #163, opened 2016-06-09
DATACMNS-868 - Repository query method support for Stream and Optional return types forces the need for Java 8 runtime.

jxblum:


olivergierke:
That's been polished and merged. Thanks!

PR #166, opened 2016-06-28
DATACMNS-531 - Add support for a simplified syntax in request paramet…

sancho21:
Added new SimpleSortHandlerMethodArgumentResolver as an alternative to SortHandlerMethodArgumentResolver implementation.

This is for simplified sort parameter with sign. I have feeling that SortDefaultUnitTests.java should be independent of the SortHandlerMethodArgumentResolver implementation.

### DIFF
--- a/src/main/java/org/springframework/data/web/SimpleSortHandlerMethodArgumentResolver.java
+++ b/src/main/java/org/springframework/data/web/SimpleSortHandlerMethodArgumentResolver.java
@@ -1,0 +1,382 @@
+/*
+ * Copyright 2013-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


olivergierke:
So what is the plan on integrating this? We can't just simply replace the current one as that would break existing applications.

Regarding the split of the unit tests for `SortDefault` — You now tied those tests to your implementation, which doesn't necessarily change the situation.

To be completely hones, I am not even sure I want to "solve" the issue, as we're basically just trading one way with the other and I am pretty sure, if we moved to the new format people that preferred the old format.

sancho21:
On https://jira.spring.io/browse/DATACMNS-531, you said that you want clear separation. So, I made a separate class for simple sort option. This way, users still can either choose the existing SortHandlerMethodArgumentResolver or the new SimpleSortHandlerMethodArgumentResolver as both classes are there for them to choose. 

Or, is it better to just add another flag to the existing class like?

jeffsegal:
+1 on this PR. It snaps nicely in and satisfies a very common pattern in APIs (+/- sort direction).

pivotal-issuemaster:
@sancho21 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=166) to manually synchronize the status of this Pull Request.

mp911de:
Thanks for your PR. Closing this one as per comment in the linked ticket as we do not intend to add additional/alternative sort formats. You can customize the format by providing your own `SortHandlerMethodArgumentResolver` bean.

PR #168, opened 2016-07-16
DATACMNS-883 - make dot path and reify path correctly for ListPath

tanapoln:


tanapoln:
duplicated with f2a6d63

PR #172, opened 2016-08-17
DATACMNS-658 - Add 'ignorecase' ordering option to SortHandlerMethodArgumentResolver

dan-nawrocki:
Updated version of https://github.com/spring-projects/spring-data-commons/pull/151 for the 1.12.x branch.

Please let me know what I can do to get similar functionality in to the next version. I understand that it's unlikely to be merged as-is, but comments on the `sort2` option in the previous pull request would be appreciated.

### DIFF
--- a/src/main/java/org/springframework/data/web/SortHandlerMethodArgumentResolver.java
+++ b/src/main/java/org/springframework/data/web/SortHandlerMethodArgumentResolver.java
@@ -195,8 +195,10 @@ public class SortHandlerMethodArgumentResolver implements HandlerMethodArgumentR
 
 	/**
 	 * Parses the given sort expressions into a {@link Sort} instance. The implementation expects the sources to be a
-	 * concatenation of Strings using the given delimiter. If the last element can be parsed into a {@link Direction} it's
-	 * considered a {@link Direction} and a simple property otherwise.
+	 * concatenation of Strings using the given delimiter. If the last element is equal to "ignorecase" (when using a
+	 * case-insensitive comparison), the sort order will be performed without respect to case. If the last element (or
+	 * the penultimate element if the last is "ignorecase") can be parsed into a {@link Direction} it's considered a
+	 * {@link Direction} and a simple property otherwise.
 	 * 
 	 * @param source will never be {@literal null}.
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@dan-nawrocki Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=172) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@dan-nawrocki Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=172)!

schauder:
Two things are currently keeping us from going forward with this:

1. When we extend the parsing part for sort parameters we should also extend the rendering part (see HateoasSortHandlerMethodArgumentResolver)

2. We would like to have something less bulky, i.e. a shorter way instead of the rather lengthy "ignorecase". Admittedly we haven't made up our mind how exactly that should look. 

Related: #207

dan-nawrocki:
@schauder thanks for the feedback. Taking your concerns and @olivergierke's thoughts on backwards compatibility, I wonder if the following proposal would be more palatable. Instead of introducing a ```sort2``` method, rely on the "shortened" sort format proposed in #207 to come up with something like:
```
sort=foo>~
```
Where the ```>``` indicates descending order and the ```~``` indicates case-insensitive searching. NULL-first could be triggered by ```{```; NULL-last triggered by ```}``` (running out of options that aren't reserved URI characters). On the downside, having a lot of parameters like this makes it much harder to remember what everything does. Something like
```
sort=foo>,bar<~{,baz>}
```
gets pretty unwieldy. Thoughts?

I won't plan on updating the HateoasSortHandlerMethodArgumentResolver since it doesn't look like this PR will be accepted as-is, but I'll keep it in mind if/when I make other updates.

mp911de:
Sorry for the long radio silence. Tweaking sort behavior via the URL is a widely discussed topic and we had to get a clear picture of the support we want to provide. We picked up this PR, rebased, merged, and polished it to `master`.

PR #173, opened 2016-08-20
DATACMNS-902 - Fix attribute name in example showing how to define a custom repository base class

icoderman:
DATAJPA-924 - fix error in Example 30 of reference documentation - attribute named repository-base-class changed to base-class


mp911de:
Thanks, that's merged and backported.

PR #177, opened 2016-09-21
DATACMNS-836 - Add components to support reactive repositories.

mp911de:
We now expose reactive interfaces to facilitate reactive repository support in store-specific modules. Spring Data modules are free to implement their reactive support using either RxJava1 or Project Reactor (Reactive Streams). We expose a set of base interfaces:
- `ReactiveCrudRepository`
- `ReactiveSortingRepository`
- `RxJava1CrudRepository`
- `RxJava1SortingRepository`

Reactive repositories provide a similar feature coverage to blocking repositories. Reactive paging is not supported.

A store module can choose either Project Reactor or RxJava 1 to implement reactive repository support. Project Reactor and RxJava types are converted in both directions allowing repositories to be composed of Project Reactor and RxJava query methods. Reactive wrapper type conversion handles wrapper type conversion at repository level. Query/implementation method selection uses multi-pass candidate selection to invoke the most appropriate method (exact arguments, convertible wrappers, assignable arguments).

We also provide `ReactiveWrappers` and `ReactiveWrapperConverters` to expose metadata about reactive types and their value multiplicity.

---

Related ticket: [DATACMNS-836](https://jira.spring.io/browse/DATACMNS-836)

Open issues:
- [x] Revisit changes
- [x] Remove touched classes without functional changes


mp911de:
`QueryExecutionConverters` uses `ReactiveAdapterRegistry` which requires Project Reactor to be available on the classpath. This breaks applications not having Project Reactor on their class path.

src/main/java/org/springframework/data/repository/reactive/RxJavaCrudRepository.java
```diff
@@ -0,0 +1,168 @@
...
+	 *
+	 * @param entityStream
+	 * @throws IllegalArgumentException in case the given {@link Publisher} is {@literal null}.
+	 */
+	Single<Void> delete(Observable<? extends T> entityStream);
+
+	/**
+	 * Deletes all entities managed by the repository.
+	 */
+	Single<Void> deleteAll();
```
mp911de: Consider `Completable` instead of `Single<Void>`

src/main/java/org/springframework/data/repository/reactive/RxJavaCrudRepository.java
```diff
@@ -0,0 +1,168 @@
...
+	Single<T> findOne(ID id);
+
+	/**
+	 * Retrieves an entity by its id supplied by a {@link Single}.
+	 *
+	 * @param id must not be {@literal null}.
+	 * @return the entity with the given id or {@literal null} if none found
+	 * @throws IllegalArgumentException if {@code id} is {@literal null}
+	 */
+	Single<T> findOne(Single<ID> id);
```
mp911de: Consider `Observable<T>` instead of `Single<T>` as the result may be absent.

olivergierke:
I pushed the changes suggested above in abac726. The use of a `ConversionService` still seems to be a bit of overkill in `WrapperConversionMatch` as I guess we could just determine the same information from `ReactiveWrappers` directly.

PR #179, opened 2016-09-28
DATACMNS-918 - Provide interfaces for Pageable and Sort method argument resolvers.

mp911de:
We now provide `PageableArgumentResolver` and `SortArgumentResolver` to enable wrapping of actual resolvers and simplify `Pageable` and `Sort` post processing.

---

Related ticket: [DATACMNS-918](https://jira.spring.io/browse/DATACMNS-918)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>1.13.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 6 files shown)


olivergierke:
Any reason we don't extend `HandlerMethodArgumentResolver` directly? With the overload of `supports(…)` we might cause issues for current custom implementations as they might need access to other parameters than we take for this method.

mp911de:
`PageableArgumentResolver` and `SortArgumentResolver` now extend `HandlerMethodArgumentResolver`.

PR #180, opened 2016-10-05
DATACMNS-925 - Improve memory consumption of Parameter and Parameters.

dreis2211:
Fixes https://jira.spring.io/browse/DATACMNS-925

### DIFF
--- a/src/main/java/org/springframework/data/repository/query/Parameters.java
+++ b/src/main/java/org/springframework/data/repository/query/Parameters.java
@@ -62,10 +62,10 @@ public abstract class Parameters<S extends Parameters<S, T>, T extends Parameter
 
 		Assert.notNull(method);
 
-		this.parameters = new ArrayList<T>();
-		this.dynamicProjectionIndex = -1;
-
 		List<Class<?>> types = Arrays.asList(method.getParameterTypes());
+
+		this.parameters = new ArrayList<T>(types.size());
+		this.dynamicProjectionIndex = -1;
 
 		for (int i = 0; i < types.size(); i++) {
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/repository/query/Parameters.java
```diff
@@ -62,11 +62,11 @@ public Parameters(Method method) {
 
 		Assert.notNull(method);
 
-		this.parameters = new ArrayList<T>();
-		this.dynamicProjectionIndex = -1;
-
 		List<Class<?>> types = Arrays.asList(method.getParameterTypes());
 
+		this.parameters = new ArrayList<T>(types.size());
```
olivergierke: AFAICS moving the code shouldn't be necessary, should it? I usually declare variables as close as possible to their usage, so I'd like to keep the old position unless there's any effect of that move.

src/main/java/org/springframework/data/repository/query/Parameters.java
```diff
@@ -98,12 +98,12 @@ public Parameters(Method method) {
 	 */
 	protected Parameters(List<T> originals) {
 
-		this.parameters = new ArrayList<T>();
-
 		int pageableIndexTemp = -1;
 		int sortIndexTemp = -1;
 		int dynamicProjectionTemp = -1;
 
+		this.parameters = new ArrayList<T>(originals.size());
```
olivergierke: Same as above. Can we keep the old position?

src/main/java/org/springframework/data/repository/query/Parameters.java
dreis2211: I need the types variable here in order to retrieve the size. If I call method.getParameters() again I have the overhead of cloning the method parameters, which I wanted to avoid. As far as I saw, you're not yet compiling against Java 8, which would have enabled me to call Method.getParameterCount().

src/main/java/org/springframework/data/repository/query/Parameters.java
dreis2211: In this case your right, this was just for "consistency" to the method above. Let me know when I should change it.

olivergierke:
That's merged, thanks!

PR #184, opened 2016-11-18
Added possibility to register custom converters of NullableWrapper

kaczynskid:
This is a re-write of `QueryExecutionConverters` to allow registration of custom converters for wrapper types. This came from the need to use JavaSlang types as return types of query methods. It was not possible to add support for JavaSlang types directly, as it's not compatible with `1.6` source level. Therefore an extension point was implemented using `ServiceLoader` pattern.

### DIFF
--- a/src/main/java/org/springframework/data/repository/util/NullableWrapperConverter.java
+++ b/src/main/java/org/springframework/data/repository/util/NullableWrapperConverter.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


olivergierke:
Would you mind creating a ticket in our JIRA first so that we can discuss the idea first before talking about implementation details? We already build with JDK 8 and thus we should be able to optionally integrate with libraries that require JDK at runtime. So we could try to get away with a plain classpath check and a simple `Converter` and thus wouldn't have to throw the baby out with the bath water.

There's no precedence in using a `ServiceLoader` in Spring Data before and I'd like to keep it this way as it just gets confusing to all of a sudden introduce a different way of configuring things (compared to plain programmatic setup, that then can be populated via Spring configuration).

kaczynskid:
Closing this as per discussion in https://jira.spring.io/browse/DATACMNS-937

PR #187, opened 2016-12-02
DATACMNS-867: Avoid allocations from Method.getParameterTypes() where…

dreis2211:
Adds to https://jira.spring.io/browse/DATACMNS-867


olivergierke:
That's merged into the feature branch. Thanks!

PR #191, opened 2017-01-16
DATACMNS-976 - AbstractEntityInformation considers Persistable.isNew().

christophstrobl:
We now consider `Persistable.isNew()` implementations when checking `isNew(entity)` on an persistent entity.


olivergierke:
Decided for a different implementation approach. See the ticket for discussion.

PR #199, opened 2017-03-07
DATACMNS-1002 - Generic class parameter not being bound to query.

mbragg02:
Marks a Parameter as bindable if it is explicitly named.

Related ticket: https://jira.spring.io/browse/DATACMNS-1002


olivergierke:
Thanks for the contribution. We're looking into alternative options to solve the issue as `@Param` is already used in completely different contexts.

PR #201, opened 2017-03-09
DATACMNS-764 - Disambiguate custom repository implementation

schauder:
Disambiguates custom implementations of repositories by comparing the bean names of the candidate implementations with the bean name of the repository.

Therefore naming the unwanted implementations or naming the repository interface and the required implementation can be used to break ambiguities.

Also, fixes the path on which test sources get generated.

### DIFF
--- a/src/main/asciidoc/repositories.adoc
+++ b/src/main/asciidoc/repositories.adoc
@@ -649,6 +649,37 @@ If you use namespace configuration, the repository infrastructure tries to autod
 
 The first configuration example will try to look up a class `com.acme.repository.UserRepositoryImpl` to act as custom repository implementation, whereas the second example will try to lookup `com.acme.repository.UserRepositoryFooBar`.
 
+===== Resolution of ambiguity
+
+If multiple implementations with matching class names get found in different packages Spring Data uses the bean names to identify the correct on to use.
+
+Given the following two custom implementations for the `UserRepository` introduced above the first implementation will get picked. It's bean name is `userRepositoryImpl` matches that of the repository interface (`userRepository`) plus the postfix `Impl`.
+
+.Resolution of amibiguous implementations
+====
+[source, java]
+----
+package com.acme.impl.one;
+
+class UserRepositoryImpl implements UserRepositoryCustom {
+
+  // Your custom implementation
+}
+----
+[source, java]
+----
+package com.acme.impl.two;
+
+@Component("specialCustomImpl")
+class UserRepositoryImpl implements UserRepositoryCustom {
+
+  // Your custom implementation
+}
+----
+====
+
+If you annotate the `UserRepository` interface with `Component("specialCustom")' the bean name plus `Impl` matches that of the second Repository and it will be picked instead of the first one.
+
 ===== Manual wiring
 
 The approach just shown works well if your custom implementation uses annotation-based configuration and autowiring only, as it will be treated as any other Spring bean. If your custom implementation bean needs special wiring, you simply declare the bean and name it after the conventions just described. The infrastructure will then refer to the manually defined bean definition by name instead of creating one itself.
--- a/src/main/asciidoc/repositories.adoc
+++ b/src/main/asciidoc/repositories.adoc
@@ -651,9 +651,9 @@ The first configuration example will try to look up a class `com.acme.repository
 
 ===== Resolution of ambiguity
 
-If multiple implementations with matching class names get found in different packages Spring Data uses the bean names to identify the correct on to use.
+If multiple implementations with matching class names get found in different packages Spring Data uses the bean names to identify the correct one to use.
 
-Given the following two custom implementations for the `UserRepository` introduced above the first implementation will get picked. It's bean name is `userRepositoryImpl` matches that of the repository interface (`userRepository`) plus the postfix `Impl`.
+Given the following two custom implementations for the `UserRepository` introduced above the first implementation will get picked. Its bean name is `userRepositoryImpl` matches that of the repository interface (`userRepository`) plus the postfix `Impl`.
 
 .Resolution of amibiguous implementations
 ====
--- a/src/main/java/org/springframework/data/repository/config/AnnotationRepositoryConfigurationSource.java
+++ b/src/main/java/org/springframework/data/repository/config/AnnotationRepositoryConfigurationSource.java
@@ -47,6 +47,7 @@ import org.springframework.util.StringUtils;
  * @author Oliver Gierke
  * @author Thomas Darimont
  * @author Peter Rietzler
+ * @author Jens Schauder
  */
 public class AnnotationRepositoryConfigurationSource extends RepositoryConfigurationSourceSupport {
 
@@ -77,7 +78,7 @@ public class AnnotationRepositoryConfigurationSource extends RepositoryConfigura
 	public AnnotationRepositoryConfigurationSource(AnnotationMetadata metadata, Class<? extends Annotation> annotation,
 			ResourceLoader resourceLoader, Environment environment) {
 
-		super(environment);
+		super(environment, resourceLoader.getClassLoader());
 
 		Assert.notNull(metadata, "Metadata must not be null!");
 		Assert.notNull(annotation, "Annotation must not be null!");
--- a/src/main/java/org/springframework/data/repository/config/CustomRepositoryImplementationDetector.java
+++ b/src/main/java/org/springframework/data/repository/config/CustomRepositoryImplementationDetector.java
@@ -19,9 +19,13 @@ package org.springframework.data.repository.config;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
@@ -37,11 +41,12 @@ import org.springframework.util.Assert;
 
 /**
  * Detects the custom implementation for a {@link org.springframework.data.repository.Repository}
- * 
+ *
  * @author Oliver Gierke
  * @author Mark Paluch
  * @author Christoph Strobl
  * @author Peter Rietzler
+ * @author Jens Schauder
  */
 @RequiredArgsConstructor
 public class CustomRepositoryImplementationDetector {
@@ -54,7 +59,7 @@ public class CustomRepositoryImplementationDetector {
 
 	/**
 	 * Tries to detect a custom implementation for a repository bean by classpath scanning.
-	 * 
+	 *
 	 * @param configuration the {@link RepositoryConfiguration} to consider.
 	 * @return the {@code AbstractBeanDefinition} of the custom implementation or {@literal null} if none found.
 	 */
@@ -62,23 +67,43 @@ public class CustomRepositoryImplementationDetector {
 
 		// TODO 2.0: Extract into dedicated interface for custom implementation lookup configuration.
 
-		return detectCustomImplementation(configuration.getImplementationClassName(), //
+		return detectCustomImplementation( //
+				configuration.getImplementationClassName(), //
+				configuration.getImplementationBeanName(), //
 				configuration.getBasePackages(), //
-				configuration.getExcludeFilters());
+				configuration.getExcludeFilters(), //
+				bd -> configuration.getConfigurationSource().generateBeanName(bd));
 	}
 
 	/**
 	 * Tries to detect a custom implementation for a repository bean by classpath scanning.
-	 * 
+	 *
 	 * @param className must not be {@literal null}.
+	 * @param beanName may be {@literal null}
 	 * @param basePackages must not be {@literal null}.
+	 * @param excludeFilters must not be {@literal null}.
+	 * @param beanNameGenerator must not be {@literal null}.
 	 * @return the {@code AbstractBeanDefinition} of the custom implementation or {@literal null} if none found.
 	 */
-	public Optional<AbstractBeanDefinition> detectCustomImplementation(String className, Iterable<String> basePackages,
-			Iterable<TypeFilter> excludeFilters) {
+	public Optional<AbstractBeanDefinition> detectCustomImplementation(String className, String beanName,
+			Iterable<String> basePackages, Iterable<TypeFilter> excludeFilters,
+			Function<BeanDefinition, String> beanNameGenerator) {
 
 		Assert.notNull(className, "ClassName must not be null!");
 		Assert.notNull(basePackages, "BasePackages must not be null!");
+
+		Set<BeanDefinition> definitions = findCandidateBeanDefinitions(className, basePackages, excludeFilters);
+
+		SelectionSet<BeanDefinition> selection = new SelectionSet<>( //
+				definitions, //
+				c -> c.isEmpty() ? null : throwAmbiguousCustomImplementationException(c) //
+		).filterIfNecessary(bd -> beanName != null && beanName.equals(beanNameGenerator.apply(bd)));
+
+		return Optional.ofNullable((AbstractBeanDefinition) selection.uniqueResult());
+	}
+
+	Set<BeanDefinition> findCandidateBeanDefinitions(String className, Iterable<String> basePackages,
+			Iterable<TypeFilter> excludeFilters) {
 
 		// Build pattern to lookup implementation class
 		Pattern pattern = Pattern.compile(".*\\." + className);
@@ -101,12 +126,15 @@ public class CustomRepositoryImplementationDetector {
 			definitions.addAll(provider.findCandidateComponents(basePackage));
 		}
 
-		if (definitions.isEmpty()) {
-			return Optional.empty();
-		}
+		return definitions;
+	}
 
-		if (definitions.size() == 1) {
-			return Optional.of((AbstractBeanDefinition) definitions.iterator().next());
+	private static AbstractBeanDefinition throwAmbiguousCustomImplementationException(
+			Collection<BeanDefinition> definitions) {
+
+		List<String> implementationClassNames = new ArrayList<String>();
+		for (BeanDefinition bean : definitions) {
+			implementationClassNames.add(bean.getBeanClassName());
 		}
 
 		throw new IllegalStateException(
--- a/src/main/java/org/springframework/data/repository/config/DefaultRepositoryConfiguration.java
+++ b/src/main/java/org/springframework/data/repository/config/DefaultRepositoryConfiguration.java
@@ -28,8 +28,9 @@ import org.springframework.util.StringUtils;
 
 /**
  * Default implementation of {@link RepositoryConfiguration}.
- * 
+ *
  * @author Oliver Gierke
+ * @author Jens Schauder
  */
 @RequiredArgsConstructor
 public class DefaultRepositoryConfiguration<T extends RepositoryConfigurationSource>
@@ -104,7 +105,8 @@ public class DefaultRepositoryConfiguration<T extends RepositoryConfigurationSou
 	 * @see org.springframework.data.repository.config.RepositoryConfiguration#getImplementationBeanName()
 	 */
 	public String getImplementationBeanName() {
-		return StringUtils.uncapitalize(getImplementationClassName());
+		return configurationSource.generateBeanName(definition)
+				+ configurationSource.getRepositoryImplementationPostfix().orElse("Impl");
 	}
 
 	/* 
--- a/src/main/java/org/springframework/data/repository/config/RepositoryBeanNameGenerator.java
+++ b/src/main/java/org/springframework/data/repository/config/RepositoryBeanNameGenerator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012 the original author or authors.
+ * Copyright 2012-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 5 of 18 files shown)


src/main/asciidoc/repositories.adoc
```diff
@@ -649,6 +649,39 @@ If you use namespace configuration, the repository infrastructure tries to autod
...
+
+Given the following two custom implementations for the `UserRepository` introduced above the first implementation will get picked. It's bean name is `userRepositoryImpl` matches that of the repository interface (`userRepository`) plus the postfix `Impl`.
+
+.Resolution of amibiguous implementations
+====
+[source, java]
+----
+package com.acme.impl.one;
+
+class UserRepositoryImpl implements UserRepositoryCustom {
```
olivergierke: I think we can even leave away the method declarations here. Should we describe how one would get the second implementation wired?

src/main/java/org/springframework/data/repository/config/AnnotationRepositoryConfigurationSource.java
```diff
@@ -87,6 +92,8 @@ public AnnotationRepositoryConfigurationSource(AnnotationMetadata metadata, Clas
 		this.configMetadata = metadata;
 		this.resourceLoader = resourceLoader;
 		this.hasExplicitFilters = hasExplicitFilters(attributes);
+
+		beanNameGenerator = new RepositoryBeanNameGenerator(resourceLoader.getClassLoader());
```
olivergierke: Qualify with `this.`.

src/main/java/org/springframework/data/repository/config/AnnotationRepositoryConfigurationSource.java
```diff
@@ -46,6 +48,7 @@
  * @author Oliver Gierke
  * @author Thomas Darimont
  * @author Peter Rietzler
+ * @author Jens Schauder
  */
 public class AnnotationRepositoryConfigurationSource extends RepositoryConfigurationSourceSupport {
```
olivergierke: Couldn't all of this go into `RepositoryConfigurationSourceSupport`?

src/main/java/org/springframework/data/repository/config/CustomRepositoryImplementationDetector.java
```diff
@@ -47,6 +50,7 @@
 	private final MetadataReaderFactory metadataReaderFactory;
 	private final Environment environment;
 	private final ResourceLoader resourceLoader;
+	private final RepositoryBeanNameGenerator nameGenerator;
```
olivergierke: I wonder if there's a way that we can avoid holding an additional generator to the one that the configuration uses. I could live with the CDI interaction becoming a bit more complicated for now as we're going to unify that into a dedicated configuration interface anyway and it would even make the requirements for such a unified interface even more explicit.

So let's say the `detectCustomImplementation(…)` overload taking a lot of parameters doesn't have any bounds for now. Let's make that as complicated as necessary to then know what needs to go into the shared config interface.

src/main/java/org/springframework/data/repository/config/CustomRepositoryImplementationDetector.java
```diff
@@ -80,23 +86,55 @@ public AbstractBeanDefinition detectCustomImplementation(RepositoryConfiguration
...
-			Iterable<TypeFilter> excludeFilters) {
+	public AbstractBeanDefinition detectCustomImplementation(String className, String beanName, Iterable<String> basePackages,
+															 Iterable<TypeFilter> excludeFilters) {
 
 		Assert.notNull(className, "ClassName must not be null!");
 		Assert.notNull(basePackages, "BasePackages must not be null!");
 
+		Set<BeanDefinition> definitions = findCandidateBeanDefinitions(className, basePackages, excludeFilters);
+
+		if (definitions.isEmpty()) {
```
olivergierke: Couldn't that logic go into `findCandidateBeanDefinitions(…)` directly?

src/main/java/org/springframework/data/repository/config/RepositoryBeanNameGenerator.java
```diff
@@ -15,56 +15,68 @@
...
 	 * Returns the type configured for the {@code repositoryInterface} property of the given bean definition. Uses a
 	 * potential {@link Class} being configured as is or tries to load a class with the given value's {@link #toString()}
 	 * representation.
-	 * 
+	 *
 	 * @param beanDefinition
 	 * @return
 	 */
-	private Class<?> getRepositoryInterfaceFrom(BeanDefinition beanDefinition) {
+	Class<?> getRepositoryInterfaceFrom(BeanDefinition beanDefinition) {
```
olivergierke: Can this stay private?

src/test/java/org/springframework/data/repository/config/DefaultRepositoryConfigurationUnitTests.java
```diff
@@ -17,29 +17,44 @@
...
 public class DefaultRepositoryConfigurationUnitTests {
 
 	@Mock RepositoryConfigurationSource source;
 
+
+	@Before
+	public void before() {
+
+		when(source.getBeanNameGenerator()).thenReturn(
+				new RepositoryBeanNameGenerator(Thread.currentThread().getContextClassLoader())
```
olivergierke: Any reason you're using this over `getClass().getClassLoader()`?

src/main/java/org/springframework/data/repository/config/CustomRepositoryImplementationDetector.java
```diff
@@ -80,23 +84,43 @@ public AbstractBeanDefinition detectCustomImplementation(RepositoryConfiguration
...
+			Iterable<TypeFilter> excludeFilters,
+			RepositoryBeanNameGenerator nameGenerator) {
 
 		Assert.notNull(className, "ClassName must not be null!");
 		Assert.notNull(basePackages, "BasePackages must not be null!");
 
+		Set<BeanDefinition> definitions = findCandidateBeanDefinitions(className, basePackages, excludeFilters);
+
+		SelectionSet<BeanDefinition> selection = new SelectionSet<>(definitions, c -> {
+			if (c.isEmpty()) {
```
olivergierke: Ternary if might allow us to condense this into a single line to avoid the block.

src/main/java/org/springframework/data/repository/config/CustomRepositoryImplementationDetector.java
```diff
@@ -80,23 +84,43 @@ public AbstractBeanDefinition detectCustomImplementation(RepositoryConfiguration
...
+				return null;
+			}
+			return throwAmbiguousCustomImplementationException(c);
+		});
+		selection.filter(bd -> beanName != null && beanName.equals(nameGenerator.generateBeanName(bd)));
```
olivergierke: Blank line before (generally alway after a block unless it's only a closing `}`).

src/main/java/org/springframework/data/repository/config/CustomRepositoryImplementationDetector.java
```diff
@@ -117,15 +141,10 @@ public AbstractBeanDefinition detectCustomImplementation(String className, Itera
 		for (String basePackage : basePackages) {
 			definitions.addAll(provider.findCandidateComponents(basePackage));
 		}
+		return definitions;
```
olivergierke: Blank line before (see above).

src/main/java/org/springframework/data/repository/config/CustomRepositoryImplementationDetector.java
```diff
@@ -117,15 +141,10 @@ public AbstractBeanDefinition detectCustomImplementation(String className, Itera
...
-		if (definitions.isEmpty()) {
-			return null;
-		}
-
-		if (definitions.size() == 1) {
-			return (AbstractBeanDefinition) definitions.iterator().next();
-		}
-
+	private static AbstractBeanDefinition throwAmbiguousCustomImplementationException(Collection<BeanDefinition> definitions) {
 		List<String> implementationClassNames = new ArrayList<String>();
```
olivergierke: Blank line before multi-line method implementations.

src/main/java/org/springframework/data/repository/config/DefaultRepositoryConfiguration.java
```diff
@@ -114,7 +115,7 @@ public String getImplementationClassName() {
 	 * @see org.springframework.data.repository.config.RepositoryConfiguration#getImplementationBeanName()
 	 */
 	public String getImplementationBeanName() {
-		return StringUtils.uncapitalize(getImplementationClassName());
+		return configurationSource.getBeanNameGenerator().generateBeanName(definition)  + getImplementationPostfix();
```
olivergierke: Assign generator to local variable. Avoid double spaces before `+`.

src/main/java/org/springframework/data/repository/config/RepositoryBeanNameGenerator.java
```diff
@@ -15,56 +15,68 @@
...
 	 * Returns the type configured for the {@code repositoryInterface} property of the given bean definition. Uses a
 	 * potential {@link Class} being configured as is or tries to load a class with the given value's {@link #toString()}
 	 * representation.
-	 * 
+	 *
 	 * @param beanDefinition
 	 * @return
 	 */
 	private Class<?> getRepositoryInterfaceFrom(BeanDefinition beanDefinition) {
+		if (beanDefinition instanceof ScannedGenericBeanDefinition) {
```
olivergierke: Blank line before multi-line method implementations.

src/main/java/org/springframework/data/repository/config/RepositoryBeanNameGenerator.java
```diff
@@ -15,56 +15,68 @@
...
+			try {
+				return ((ScannedGenericBeanDefinition) beanDefinition).resolveBeanClass(beanClassLoader);
+			} catch (ClassNotFoundException e) {
+				throw new IllegalStateException("Could not resolve bean class.", e);
+			}
+		} else {
```
olivergierke: We don't need an else-block here as the first branch always yields a result, right?

src/main/java/org/springframework/data/repository/config/RepositoryBeanNameGenerator.java
```diff
@@ -15,56 +15,68 @@
...
+			return getRepositoryInterfaceFromFactory(beanDefinition);
+		}
+	}
 
+	private Class<?> getRepositoryInterfaceFromFactory(BeanDefinition beanDefinition) {
 		Object value = beanDefinition.getConstructorArgumentValues().getArgumentValue(0, Class.class).getValue();
```
olivergierke: Blank line before multi-line method implementations.

src/main/java/org/springframework/data/repository/config/RepositoryConfigurationDelegate.java
```diff
@@ -133,7 +128,7 @@ private static Environment defaultEnvironment(Environment environment, ResourceL
 			}
 
 			AbstractBeanDefinition beanDefinition = definitionBuilder.getBeanDefinition();
-			String beanName = beanNameGenerator.generateBeanName(beanDefinition, registry);
+			String beanName = configurationSource.getBeanNameGenerator().generateBeanName(beanDefinition);
```
olivergierke: Looks like the pattern "get generator to then generate a name" is a very common pattern (seen above as well). Does it maybe make sense to — instead of exposing the generator — introduce a shortcut method on the configuration source to generate a bean name for a definition?

src/main/java/org/springframework/data/repository/config/RepositoryConfigurationSourceSupport.java
```diff
@@ -28,26 +28,32 @@
...
 	 * Creates a new {@link RepositoryConfigurationSourceSupport} with the given environment.
-	 * 
+	 *
 	 * @param environment must not be {@literal null}.
+	 * @param classLoader
 	 */
-	public RepositoryConfigurationSourceSupport(Environment environment) {
+	public RepositoryConfigurationSourceSupport(Environment environment, ClassLoader classLoader) {
 
 		Assert.notNull(environment, "Environment must not be null!");
```
olivergierke: I'd rather put a blank line after the assert and remove the one below to group the assignments.

src/main/java/org/springframework/data/repository/config/RepositoryConfigurationSourceSupport.java
```diff
@@ -28,26 +28,32 @@
...
 	private final Environment environment;
 
+	private final RepositoryBeanNameGenerator beanNameGenerator;
+
 	/**
 	 * Creates a new {@link RepositoryConfigurationSourceSupport} with the given environment.
-	 * 
+	 *
 	 * @param environment must not be {@literal null}.
+	 * @param classLoader
```
olivergierke: Nullability requirements? Missing assertion?

src/main/java/org/springframework/data/repository/config/RepositoryConfigurationSourceSupport.java
```diff
@@ -98,10 +104,19 @@ public RepositoryConfigurationSourceSupport(Environment environment) {
...
 	 * classes.
-	 * 
+	 *
 	 * @return {@literal true} if the container should look for nested repository interface definitions.
 	 */
 	public boolean shouldConsiderNestedRepositories() {
 		return false;
 	}
+
+	/*
```
olivergierke: Formatting seems off here. Caused by spaces maybe?

src/main/java/org/springframework/data/repository/config/SelectionSet.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Allows filtering of a collection in order to select a unique element. Once a unique element is found all further
+ * filters are ignored.
+ *
+ * @author Jens Schauder
+ */
+class SelectionSet<T> {
+
+	private Collection<T> collection;
```
olivergierke: Final?

src/main/java/org/springframework/data/repository/config/SelectionSet.java
```diff
@@ -0,0 +1,101 @@
...
+	 *
+	 * @param predicate To be used for filtering.
+	 */
+	void filter(Predicate<T> predicate) {
+
+		if (uniqueResult != null) {
+			return;
+		}
+
+		collection = collection.stream().filter(predicate).collect(Collectors.toList());
```
olivergierke: Shouldn't we rather return a new instance?

src/main/java/org/springframework/data/repository/config/SelectionSet.java
```diff
@@ -0,0 +1,101 @@
...
+
+		return uniqueResult != null ? uniqueResult : fallback.apply(collection);
+	}
+
+	/**
+	 * Filters the collection with the predicate if there are still more then one elements in the collection.
+	 *
+	 * @param predicate To be used for filtering.
+	 */
+	void filter(Predicate<T> predicate) {
```
olivergierke: Should this be `filterIfNecessary(…)` instead? With `filter(…)` I don't necessarily expect the predicate not to be applied in case there already is a unique result.

src/test/java/org/springframework/data/repository/cdi/CdiRepositoryBeanUnitTests.java
```diff
@@ -119,6 +124,36 @@ public void createsPassivationId() {
 		assertThat(bean.getId(), is(PASSIVATION_ID));
 	}
 
+	@Test // DATACMNS-764
+	public void passesCorrectBeanNameToTheImplementationDetector() {
+
+		CustomRepositoryImplementationDetector detector = mock(CustomRepositoryImplementationDetector.class);
+
+		CdiRepositoryBean<SampleRepository> bean = new CdiRepositoryBean<SampleRepository>(
```
olivergierke: `//` needed.

src/test/java/org/springframework/data/repository/config/DefaultRepositoryConfigurationUnitTests.java
```diff
@@ -17,29 +17,44 @@
...
 @RunWith(MockitoJUnitRunner.class)
 public class DefaultRepositoryConfigurationUnitTests {
 
 	@Mock RepositoryConfigurationSource source;
 
+
+	@Before
+	public void before() {
+
+		when(source.getBeanNameGenerator()).thenReturn(
```
olivergierke: Extract variable and remove breaks.

src/test/java/org/springframework/data/repository/config/DefaultRepositoryConfigurationUnitTests.java
```diff
@@ -48,4 +63,15 @@ public void supportsBasicConfiguration() {
 		assertThat(configuration.getQueryLookupStrategyKey(), is((Object) Key.CREATE_IF_NOT_FOUND));
 		assertThat(configuration.isLazyInit(), is(false));
 	}
+
+	private RootBeanDefinition createBeanDefinition() {
```
olivergierke: Static?

src/test/java/org/springframework/data/repository/config/SelectionSetUnitTests.java
```diff
@@ -0,0 +1,64 @@
...
+
+	@Test // DATACMNS-764
+	public void ignoresFilterWhenResultIsAlreadyUnique() {
+
+		SelectionSet<String> selection = new SelectionSet<>(asList("one"));
+		selection.filter(s -> s.contains("w"));
+
+		assertEquals("one", selection.uniqueResult());
+	}
+}
```
olivergierke: Line break.

src/main/java/org/springframework/data/repository/config/SelectionSet.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * Allows filtering of a collection in order to select a unique element. Once a unique element is found all further
+ * filters are ignored.
+ *
+ * @author Jens Schauder
+ */
+class SelectionSet<T> {
```
olivergierke: I am wondering whether we shouldn't let `uniqueResult()` return an `Optional<T>` so that the client is forced to deal with the absence of the a value. Currently in that case we just throw an `IllegalStateException` saying "More then one element in collection." which doesn't give a lot of context. I'd rather let the client — who has that context — be forced to mitigate between the `Optional<T>` and `T` as it know what exactly the absence means and can craft a way more meaningful exception. With the current state, the client basically assumes the call would produce something reasonable and moves on.

src/main/java/org/springframework/data/repository/config/SelectionSet.java
```diff
@@ -0,0 +1,101 @@
...
+	void filter(Predicate<T> predicate) {
+
+		if (uniqueResult != null) {
+			return;
+		}
+
+		collection = collection.stream().filter(predicate).collect(Collectors.toList());
+	}
+
+	private static <S> Function<Collection<S>, S> defaultCallback() {
```
olivergierke: Should this be `default*F*allback()`?

src/main/java/org/springframework/data/repository/config/SelectionSet.java
```diff
@@ -0,0 +1,101 @@
...
+
+		return c -> {
```
olivergierke: Ternary if?

src/main/java/org/springframework/data/repository/config/SelectionSet.java
```diff
@@ -0,0 +1,101 @@
...
+				return null;
+			} else {
+				throw new IllegalStateException("More then one element in collection.");
+			}
+		};
+	}
+
+	private T findUniqueResult() {
+
+		if (collection.size() == 1) {
```
olivergierke: Ternary if?

src/main/java/org/springframework/data/repository/config/RepositoryConfigurationSource.java
```diff
@@ -118,4 +120,22 @@
 	 * @return must not be {@literal null}.
 	 */
 	Iterable<TypeFilter> getExcludeFilters();
+
+	/**
+	 * Returns the {@link BeanNameGenerator}
+	 * @return must not be {@literal null}
+	 * @since 2.0
+	 */
+	RepositoryBeanNameGenerator getBeanNameGenerator();
```
olivergierke: We shouldn't need this method anymore with `generateBeanName(…)` introduced, right? If that's the case, I'd rather drop the default implementation and move the code into the implementation base class.

src/main/java/org/springframework/data/repository/config/RepositoryConfigurationSourceSupport.java
```diff
@@ -28,26 +28,33 @@
...
  * @author Oliver Gierke
  * @author Thomas Darimont
  * @author Peter Rietzler
+ * @author Jens Schauder
  */
 public abstract class RepositoryConfigurationSourceSupport implements RepositoryConfigurationSource {
 
 	protected static final String DEFAULT_REPOSITORY_IMPL_POSTFIX = "Impl";
 
 	private final Environment environment;
```
olivergierke: Remove blank line.

src/test/java/org/springframework/data/repository/config/CustomRepositoryImplementationDetectorUnitTests.java
```diff
@@ -0,0 +1,114 @@
...
+		doReturn(new HashSet<>(asList(wrongBeanDefinition, expectedBeanDefinition)))
+				.when(detector)
+				.findCandidateBeanDefinitions(anyString(), anyListOf(String.class), anyListOf(TypeFilter.class));
+
+		AbstractBeanDefinition beanDefinition = detector
+				.detectCustomImplementation("className", "expected", emptyList(), emptyList(), nameGenerator);
+
+		System.out.println(beanDefinition);
+	}
+}
```
olivergierke: Line break.

src/main/java/org/springframework/data/repository/config/SelectionSet.java
```diff
@@ -0,0 +1,97 @@
...
+
+		return uniqueResult != null ? uniqueResult : fallback.apply(collection);
+	}
+
+	/**
+	 * Filters the collection with the predicate if there are still more then one elements in the collection.
+	 *
+	 * @param predicate To be used for filtering.
+	 */
+	SelectionSet<T> filter(Predicate<T> predicate) {
```
olivergierke: let's rename this to `filterIfNecessary(…)` to make obvious that the given `Predicate` might not be applied.

src/main/java/org/springframework/data/repository/config/RepositoryConfigurationSourceSupport.java
schauder: done

src/main/java/org/springframework/data/repository/config/SelectionSet.java
schauder: ok

src/test/java/org/springframework/data/repository/config/CustomRepositoryImplementationDetectorUnitTests.java
schauder: done (and remove the sout as well)

src/test/java/org/springframework/data/repository/config/SelectionSetUnitTests.java
```diff
@@ -0,0 +1,62 @@
...
+	@Test(expected = NullPointerException.class) // DATACMNS-764
+	public void throwsCustomExceptionWhenConfigured() {
+
+		new SelectionSet<>(asList("one", "two"), c -> {
+			throw new NullPointerException();
+		}).uniqueResult();
+	}
+
+	@Test // DATACMNS-764
+	public void useseFallbackWhenConfigured() {
```
xhaggi: typo here?

src/main/asciidoc/repositories.adoc
```diff
@@ -649,6 +649,37 @@ If you use namespace configuration, the repository infrastructure tries to autod
 
 The first configuration example will try to look up a class `com.acme.repository.UserRepositoryImpl` to act as custom repository implementation, whereas the second example will try to lookup `com.acme.repository.UserRepositoryFooBar`.
 
+===== Resolution of ambiguity
+
+If multiple implementations with matching class names get found in different packages Spring Data uses the bean names to identify the correct on to use.
```
xhaggi: Typo: correct **one** to use.

src/main/asciidoc/repositories.adoc
```diff
@@ -649,6 +649,37 @@ If you use namespace configuration, the repository infrastructure tries to autod
...
+
+Given the following two custom implementations for the `UserRepository` introduced above the first implementation will get picked. It's bean name is `userRepositoryImpl` matches that of the repository interface (`userRepository`) plus the postfix `Impl`.
```
xhaggi: Type: **Its** bean name

src/main/java/org/springframework/data/repository/config/SelectionSet.java
```diff
@@ -0,0 +1,95 @@
...
+		this.collection = collection;
+		this.fallback = fallback;
+	}
+
+	/**
+	 * Get the result after applying all filters sequentially until a unique result is identified. If no unique result can
+	 * be identified the callback passed in at the constructor gets called and its return value becomes the return value
+	 * of this method.
+	 *
+	 * @return a unique result, or the result of the callback provided in the constructor.
```
xhaggi: The javadoc is a bit confusing. The method does not apply any filter, because there is no filter applied if only one item exists within the collection. And you talk about a callback, but in parameter in the Ctor is named fallback.

schauder:
Rebased the branch to incorporate changes on master: 
* Usage of Optional
* Update of Mockito version

Also fixed the issues pointed out by @xhaggi

PR #203, opened 2017-03-10
DATACMNS-1007 - Is empty is not empty

BigMichi1:
add IsEmpty and IsNotEmpty keyword parsing


olivergierke:
Polished and merged.

PR #205, opened 2017-03-28
Add support for providing default querydsl bindings.

marceloverdijk:
I would like to be able to provide default querydsl bindings for all `@QuerydslPredicate` being used.

You could think about excluding unlisted properties by default or add default behaviour for certain types.

Of course it should still be possible to provide a custom `QuerydslBinderCustomizer` via the `@QuerydslPredicate` but it should also apply the default bindings.

Something like this:

```
public class QuerydslWebConfig extends QuerydslWebConfiguration {

    @Bean
    public QuerydslBindingsFactory querydslBindingsFactory() {
        return new QuerydslBindingsFactory(SimpleEntityPathResolver.INSTANCE, defaultBindings());
    }

    public QuerydslBinderCustomizer<?> defaultBindings() {
        return new QuerydslBinderCustomizer<?>() {
            
            @Override
            public void customize(final QuerydslBindings bindings, final EntityPath root) {
                bindings.excludeUnlistedProperties(true);
                bindings.bind(String.class).first((StringPath path, String value) -> {
                    // default logic for String type.
                });
            }
        };
    }
}

```

https://github.com/spring-projects/spring-data-commons/pull/202


marceloverdijk:
Created PR against wrong branch. Closing this one. Sorry.

PR #207, opened 2017-04-09
DATACMNS-531 - More concise sort syntax for sort request parameters.

thomasdarimont:
Sort orders provided via `sort` request parameter can now be defined
in a more concise way by prefixing a property with `+` and `-` which
marks it to be sorted in `ASC` or `DESC` order respectively.
If a property is not prefixed the default order resolution applies.

A parameter like `sort=-field1,+field2,field3,DESC`
yields the sort order `field1 DESC, field2 ASC, field3 DESC`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.0.0.BUILD-SNAPSHOT</version>
+	<version>2.0.0.DATACMNS-531-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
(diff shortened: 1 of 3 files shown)


thomasdarimont:
Meeh...

it seems that the `+` sign is eaten by the the servlet container when 
accessing the `sort` request parameter :/
One either would have to URL encode those URLs, manually parse the QueryString 
or just use other characters.

```java
@RestController
class FooController {
	
	@GetMapping("/")
	String sortTest(Sort sort){
		return sort.toString();
	}
}
```

http://localhost:8080/?sort=-field1,+field2,field3,DESC yields 
```
field1: DESC,field2: DESC,field3: DESC
```

http://localhost:8080/?sort=-field1,%2bfield2,field3,DESC yields
```
field1: DESC,field2: ASC,field3: DESC
```

Alternative characters `<` and `>`:
```
< == + == ASC
> == - == DESC
```

Usage:
```
>field1,<field2,field3,DESC
```

would yield:
```
field1: DESC,field2: ASC,field3: DESC
```

mp911de:
We discussed this issue with our team and came to the conclusion that we do not want to change the format nor introduce an alternative one. `SortHandlerMethodArgumentResolver` beans can be overridden to customize the format. Unless there's a good RFC or standard that emerges, we want to stick with the current form.

PR #192, opened 2017-01-22
DATACMNS-822 - Provide customizers for default pageable and sort resolvers.

vpavic:
This is a proposed solution for [DATACMNS-822](https://jira.spring.io/browse/DATACMNS-822). Also see spring-projects/spring-boot#7775.

- [x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [x] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATACMNS).
- [x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [x] You submit test cases (unit or integration tests) that back your changes.
- [x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

PR #208, opened 2017-04-10
DATACMNS-822 - Provide customizers for default pageable and sort resolvers.

schauder:
This is a proposed solution for [DATACMNS-822](https://jira.spring.io/browse/DATACMNS-822). Also, see spring-projects/spring-boot#7775.

This replaces #192. Rebased it on the new master and changed Hamcrest to AssertJ

### DIFF
--- a/src/main/java/org/springframework/data/web/config/PageableHandlerMethodArgumentResolverCustomizer.java
+++ b/src/main/java/org/springframework/data/web/config/PageableHandlerMethodArgumentResolverCustomizer.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


src/main/java/org/springframework/data/web/config/PageableHandlerMethodArgumentResolverCustomizer.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.springframework.data.web.PageableHandlerMethodArgumentResolver;
+
+/**
+ * Callback interface that can be implemented by beans wishing to customize the
+ * {@link PageableHandlerMethodArgumentResolver} configuration.
+ *
+ * @author Vedran Pavic
+ */
+public interface PageableHandlerMethodArgumentResolverCustomizer {
```
christophstrobl: since this is going to be a feature for the `Kay` release we could add `@FunctionalInterface` here. Not sure if extending `Consumer<PageableHandlerMethodArgumentResolver>` and removing `customize` makes sense here.
Also I'd prefer a naming like `PageableHandlerMethodArgumentResolverConfigurationAdapter` over `...*Customizer`.

src/main/java/org/springframework/data/web/config/SpringDataWebConfiguration.java
```diff
@@ -54,8 +57,13 @@
 	private final ApplicationContext context;
 	private final ObjectFactory<ConversionService> conversionService;
 
-	public SpringDataWebConfiguration(ApplicationContext context,
-			@Qualifier("mvcConversionService") ObjectFactory<ConversionService> conversionService) {
+	@Autowired private Optional<PageableHandlerMethodArgumentResolverCustomizer> pageableResolverCustomizer;
```
christophstrobl: can we do without the field injection here?

PR #209, opened 2017-04-20
DATACMNS-1034 - Introduced API to easily register bidirectional, type-based converters.

olivergierke:
Introduced `ConverterBuilder` which exposes API to register Spring `GenericConverter`s for the use with a store module's `CustomConversions`. This allows simple registration of such converters using Java 8 lambdas.

```java
ConverterAware converters = ConverterBuilder
  .reading(String.class, Long.class, it -> Long.valueOf(it))
  .andWriting(it -> Object::toString);
```

The setup can also be done from the reading side which would just need the method invocations inverted. The resulting `ConverterAware` will expose the registered converters so that they can be easily handed to a `CustomConversions` instance. Partial creation of either reading or writing converters is possible, too with the returned instance then only exposing one of the two converters.

### DIFF
--- a/src/main/java/org/springframework/data/convert/ConverterBuilder.java
+++ b/src/main/java/org/springframework/data/convert/ConverterBuilder.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.convert;
+
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Function;
+
+import org.springframework.core.convert.converter.GenericConverter;
+import org.springframework.core.convert.converter.GenericConverter.ConvertiblePair;
+import org.springframework.util.Assert;
+
+/**
+ * API to easily set up {@link GenericConverter} instances using Java 8 lambdas, mostly in bidirectional fashion for
+ * easy registration as custom type converters of the SPring Data mapping subsystem. The registration starts either with
+ * the definition of a reading or writing converter that can then be completed
+ * 
+ * @author Oliver Gierke
+ * @since 2.0
+ * @see #reading(Class, Class, Function)
+ * @see #writing(Class, Class, Function)
+ * @soundtrack John Mayer - Moving On and Getting Over (The Search for Everything)
+ */
+public interface ConverterBuilder {
+
+	/**
+	 * Creates a new {@link ReadingConverterBuilder} to produce a converter to read values of the given source (the store
+	 * type) into the given target (the domain type).
+	 * 
+	 * @param source must not be {@literal null}.
+	 * @param target must not be {@literal null}.
+	 * @param function must not be {@literal null}.
+	 * @return
+	 */
+	static <S, T> ReadingConverterBuilder<T, S> reading(Class<T> source, Class<S> target,
+			Function<? super T, ? extends S> function) {
+
+		Assert.notNull(source, "Source type must not be null!");
+		Assert.notNull(target, "Target type must not be null!");
+		Assert.notNull(function, "Conversion function must not be null!");
+
+		return new DefaultConverterBuilder<>(new ConvertiblePair(source, target), Optional.empty(), Optional.of(function));
+	}
+
+	/**
+	 * Creates a new {@link WritingConverterBuilder} to produce a converter to write values of the given source (the
+	 * domain type type) into the given target (the store type).
+	 * 
+	 * @param source must not be {@literal null}.
+	 * @param target must not be {@literal null}.
+	 * @param function must not be {@literal null}.
+	 * @return
+	 */
+	static <S, T> WritingConverterBuilder<S, T> writing(Class<S> source, Class<T> target,
+			Function<? super S, ? extends T> function) {
+
+		Assert.notNull(source, "Source type must not be null!");
+		Assert.notNull(target, "Target type must not be null!");
+		Assert.notNull(function, "Conversion function must not be null!");
+
+		return new DefaultConverterBuilder<>(new ConvertiblePair(target, source), Optional.of(function), Optional.empty());
+	}
+
+	/**
+	 * Returns all {@link GenericConverter} instances to be registered for the current {@link ConverterBuilder}.
+	 * 
+	 * @return
+	 */
+	Set<GenericConverter> getConverters();
+
+	/**
+	 * Exposes a writing converter.
+	 *
+	 * @author Oliver Gierke
+	 * @since 2.0
+	 */
+	interface WritingConverterAware {
+
+		/**
+		 * Returns the writing converter already created.
+		 * 
+		 * @return
+		 */
+		GenericConverter getWritingConverter();
+	}
+
+	/**
+	 * Exposes a reading converter.
+	 *
+	 * @author Oliver Gierke
+	 * @since 2.0
+	 */
+	interface ReadingConverterAware {
+
+		/**
+		 * Returns the reading converter already created.
+		 * 
+		 * @return
+		 */
+		GenericConverter getReadingConverter();
+	}
+
+	/**
+	 * Interface to represent an intermediate setup step of {@link ConverterAware} defining a reading converter first.
+	 *
+	 * @author Oliver Gierke
+	 * @since 2.0
+	 */
+	interface ReadingConverterBuilder<T, S> extends ConverterBuilder, ReadingConverterAware {
+
+		/**
+		 * Creates a new {@link ConverterAware} by registering the given {@link Function} to add a write converter.
+		 * 
+		 * @param function must not be {@literal null}.
+		 * @return
+		 */
+		ConverterAware andWriting(Function<? super S, ? extends T> function);
+	}
+
+	/**
+	 * Interface to represent an intermediate setup step of {@link ConverterAware} defining a writing converter first.
+	 *
+	 * @author Oliver Gierke
+	 * @since 2.0
+	 */
+	interface WritingConverterBuilder<S, T> extends ConverterBuilder, WritingConverterAware {
+
+		/**
+		 * Creates a new {@link ConverterAware} by registering the given {@link Function} to add a write converter.
+		 * 
+		 * @param function must not be {@literal null}.
+		 * @return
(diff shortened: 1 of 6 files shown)


readme.md
```diff
@@ -31,7 +31,7 @@ This README as well as the [reference documentation](http://docs.spring.io/sprin
 
 The main project [website](http://projects.spring.io/spring-data/) contains links to basic project information such as source code, JavaDocs, Issue tracking, etc.
 
-For more detailed questions, please refer to [spring-data on stackoverflow](http://stackoverflow.com/questions/tagged/spring-data). If you are new to Spring as well as to Spring Data, look for information about [Spring projects](https://spring.io/projects). 
+For more detailed questions, please refer to [spring-data on stackoverflow](http://stackoverflow.com/questions/tagged/spring-data). If you are new to Spring as well as to Spring Data, look for information about [Spring projects](https://spring.io/projects).
```
gregturn: Not the place to suggest moving to asciidoctor?

src/main/java/org/springframework/data/convert/ConverterBuilder.java
```diff
@@ -0,0 +1,157 @@
...
+	 * Creates a new {@link ReadingConverterBuilder} to produce a converter to read values of the given source (the store
+	 * type) into the given target (the domain type).
+	 * 
+	 * @param source must not be {@literal null}.
+	 * @param target must not be {@literal null}.
+	 * @param function must not be {@literal null}.
+	 * @return
+	 */
+	static <S, T> ReadingConverterBuilder<T, S> reading(Class<T> source, Class<S> target,
+			Function<? super T, ? extends S> function) {
```
gregturn: `Class<T> source` looks backward (S is source, right?)

src/main/java/org/springframework/data/convert/ConverterBuilder.java
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	/**
+	 * Interface to represent an intermediate setup step of {@link ConverterAware} defining a reading converter first.
+	 *
+	 * @author Oliver Gierke
+	 * @since 2.0
+	 */
+	interface ReadingConverterBuilder<T, S> extends ConverterBuilder, ReadingConverterAware {
+
```
gregturn: Should this be `<S, T>`?

src/main/java/org/springframework/data/convert/ConverterBuilder.java
```diff
@@ -0,0 +1,157 @@
...
+		 * 
+		 * @param function must not be {@literal null}.
+		 * @return
+		 */
+		ConverterAware andReading(Function<? super T, ? extends S> function);
+	}
+
+	/**
+	 * A {@link ConverterBuilder} ware of both a reading and writing converter.
+	 *
```
gregturn: *aware*

src/main/java/org/springframework/data/convert/DefaultConverterBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	/*
+	 * (non-Javadoc)
+	 * @see org.springframework.data.convert.ConverterBuilder#getConverters()
+	 */
+	@Override
+	public Set<GenericConverter> getConverters() {
+		return Optionals.toStream(getOptionalReadingConverter(), getOptionalWritingConverter()).collect(Collectors.toSet());
+	}
```
gregturn: Maybe it's just me, but I almost prefer reading this Java 8 stream as..

```
return Optionals
    .toStream(...)
    .collect(...)
```

Even if this results in an extra blank line to comply. ¯\_(ツ)_/¯

src/main/java/org/springframework/data/convert/DefaultConverterBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+		@Override
+		public Set<ConvertiblePair> getConvertibleTypes() {
+			return Collections.singleton(convertiblePair);
+		}
+
+		@WritingConverter
+		private static class Writing<S, T> extends ConfigurableGenericConverter<S, T> {
+
+			public Writing(ConvertiblePair convertiblePair, Function<? super S, ? extends T> function) {
+				super(convertiblePair, function);
```
gregturn: Is public visibility modifier required?

src/main/java/org/springframework/data/convert/DefaultConverterBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+			}
+		}
+
+		@ReadingConverter
+		private static class Reading<S, T> extends ConfigurableGenericConverter<S, T> {
+
+			public Reading(ConvertiblePair convertiblePair, Function<? super S, ? extends T> function) {
+				super(convertiblePair, function);
```
gregturn: Public required?

src/main/java/org/springframework/data/web/PagedResourcesAssembler.java
```diff
@@ -129,23 +130,41 @@ public void setForceFirstAndLastRels(boolean forceFirstAndLastRels) {
...
 
 		Assert.notNull(link, "Link must not be null!");
-		return createResource(page, assembler, link);
+
+		return createResource(page, assembler, Optional.of(link));
+	}
+
+	/**
+	 * Creates a {@link PagedResources} with an empt collection {@link EmbeddedWrapper} for the given domain type.
+	 *
```
gregturn: *empty*

src/main/java/org/springframework/data/web/PagedResourcesAssembler.java
```diff
@@ -129,23 +130,41 @@ public void setForceFirstAndLastRels(boolean forceFirstAndLastRels) {
...
 	}
 
 	/**
 	 * Creates a {@link PagedResources} with an empt collection {@link EmbeddedWrapper} for the given domain type.
 	 * 
 	 * @param page must not be {@literal null}, content must be empty.
 	 * @param type must not be {@literal null}.
-	 * @param link can be {@literal null}.
+	 * @param link must not be {@literal null}.
 	 * @return
```
gregturn: Is this actually the place for specify this, given the assertion isn't here?

src/test/java/org/springframework/data/convert/ConverterBuilderUnitTests.java
```diff
@@ -0,0 +1,89 @@
...
+		assertThat(converter.getConvertibleTypes())
+				.containsExactly(new ConvertiblePair(source.getClass(), target.getClass()));
+		assertThat(converter.convert(source, TypeDescriptor.forObject(source), TypeDescriptor.forObject(target)))
+				.isEqualTo(target);
+	}
+
+	private static void assertOnlyConverter(ConverterBuilder builder, Supplier<GenericConverter> supplier) {
+		assertThat(builder.getConverters()).containsExactly(supplier.get());
+	}
+}
```
gregturn: This test case is simply beautiful.

olivergierke:
Somehow some bogus commits slipped into this one. Removing.

src/main/java/org/springframework/data/convert/ConverterBuilder.java
olivergierke: Agreed, I flipped those.

olivergierke:
Tweaked code according to @gregturn's suggestions.

src/main/java/org/springframework/data/convert/ConverterBuilder.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Set;
+import java.util.function.Function;
+
+import org.springframework.core.convert.converter.GenericConverter;
+import org.springframework.core.convert.converter.GenericConverter.ConvertiblePair;
+import org.springframework.util.Assert;
+
+/**
+ * API to easily set up {@link GenericConverter} instances using Java 8 lambdas, mostly in bidirectional fashion for
+ * easy registration as custom type converters of the SPring Data mapping subsystem. The registration starts either with
```
mp911de: s/SPring/Spring

src/main/java/org/springframework/data/convert/ConverterBuilder.java
```diff
@@ -0,0 +1,157 @@
...
+ * the definition of a reading or writing converter that can then be completed
```
mp911de: Add dot at the end.

src/main/java/org/springframework/data/convert/ConverterBuilder.java
```diff
@@ -0,0 +1,157 @@
...
+		Assert.notNull(source, "Source type must not be null!");
+		Assert.notNull(target, "Target type must not be null!");
+		Assert.notNull(function, "Conversion function must not be null!");
+
+		return new DefaultConverterBuilder<>(new ConvertiblePair(source, target), Optional.empty(), Optional.of(function));
+	}
+
+	/**
+	 * Creates a new {@link WritingConverterBuilder} to produce a converter to write values of the given source (the
+	 * domain type type) into the given target (the store type).
```
mp911de: there's 2x type in domain type type

src/main/java/org/springframework/data/convert/ConverterBuilder.java
```diff
@@ -0,0 +1,157 @@
...
+	 */
+	Set<GenericConverter> getConverters();
+
+	/**
+	 * Exposes a writing converter.
+	 *
+	 * @author Oliver Gierke
+	 * @since 2.0
+	 */
+	public interface WritingConverterAware {
```
olivergierke: Interesting, I though at some point, Eclipse presented them as package protected so I thought the default of public only applied to methods. Rolling this back then.

src/main/java/org/springframework/data/convert/ConverterBuilder.java
```diff
@@ -0,0 +1,157 @@
...
+		ConverterAware andReading(Function<? super T, ? extends S> function);
+	}
+
+	/**
+	 * A {@link ConverterBuilder} aware of both a reading and writing converter.
+	 *
+	 * @author Oliver Gierke
+	 * @since 2.0
+	 */
+	public interface ConverterAware extends ConverterBuilder, ReadingConverterAware, WritingConverterAware {}
```
olivergierke: Naming suggestions? ;)

src/main/java/org/springframework/data/convert/CustomConversions.java
```diff
@@ -149,33 +153,46 @@ public boolean isSimpleType(Class<?> type) {
...
+	 * @param candidate must not be {@literal null}.
+	 * @param conversionService must not be {@literal null}.
+	 */
+	private void registerConverterIn(Object candidate, ConverterRegistry conversionService) {
 
-			if (it instanceof Converter) {
-				conversionService.addConverter(Converter.class.cast(it));
-				added = true;
-			}
+		boolean added = false;
```
olivergierke: At a quick glance, yes. Not at a second. The approach used allows the instance to implement multiple interfaces and thus obtains all `ConvertibleType` instances that implementing multiple ones would expose. An early return would stop that process after the first step.

Not sure this is still needed as I am unsure how likely this is in the first place. I just kept it around to preserve behavioural consistency to the old `CustomConversions`. We could think about removing that functionality, though.

PR #210, opened 2017-04-20
DATACMNS-1035 - Moved CustomConversions to Spring Data Commons.

olivergierke:
Extracted CustomConversions — whose code has largely been duplicated between the MongoDB, Couchbase and Cassandra modules — into Spring Data Commons. Store-specific extensions can now be contributed via a StoreConversions value type that carries both, store-specific default converters as well as a store-specific SimpleTypeHolder to augment the default list of simple types.

Removed SimpleTypeHolders public default constructor in favour of a protected one and a static DEFAULT instance for plain references.

### DIFF
--- a/src/main/java/org/springframework/data/convert/CustomConversions.java
+++ b/src/main/java/org/springframework/data/convert/CustomConversions.java
@@ -1,0 +1,489 @@
+/*
+ * Copyright 2011-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.convert;
+
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+import lombok.Value;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.springframework.core.GenericTypeResolver;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.core.convert.converter.ConverterFactory;
+import org.springframework.core.convert.converter.GenericConverter;
+import org.springframework.core.convert.converter.GenericConverter.ConvertiblePair;
+import org.springframework.core.convert.support.GenericConversionService;
+import org.springframework.data.mapping.model.SimpleTypeHolder;
+import org.springframework.data.util.Optionals;
+import org.springframework.data.util.Streamable;
+import org.springframework.util.Assert;
+
+/**
+ * Value object to capture custom conversion. That is essentially a {@link List} of converters and some additional logic
+ * around them. The converters build up two sets of types which store-specific basic types can be converted into and
+ * from. These types will be considered simple ones (which means they neither need deeper inspection nor nested
+ * conversion. Thus the {@link CustomConversions} also act as factory for {@link SimpleTypeHolder} .
+ * 
+ * @author Oliver Gierke
+ * @author Thomas Darimont
+ * @author Christoph Strobl
+ * @author Mark Paluch
+ * @since 2.0
+ */
+@Slf4j
+public class CustomConversions {
+
+	private static final String READ_CONVERTER_NOT_SIMPLE = "Registering converter from %s to %s as reading converter although it doesn't convert from a store-supported type! You might wanna check you annotation setup at the converter implementation.";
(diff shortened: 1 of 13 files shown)


src/main/java/org/springframework/data/convert/CustomConversions.java
```diff
@@ -0,0 +1,489 @@
...
+		 * Creates a new {@link StoreConversions} for the given store-specific {@link SimpleTypeHolder} and the given
+		 * converters.
+		 * 
+		 * @param storeTypeHolder must not be {@literal null}.
+		 * @param converters must not be {@literal null}.
+		 * @return
+		 */
+		public static StoreConversions of(SimpleTypeHolder storeTypeHolder, Object... converters) {
+
+			Assert.notNull(storeTypeHolder, "SimpleTypeHolder must not be null!");
```
mp911de: These two overloads make it quite easy to use by accident the wrong method. Should check here that `converters` isn't a single `Collection` element?

src/main/java/org/springframework/data/convert/CustomConversions.java
```diff
@@ -0,0 +1,489 @@
...
+		this.rawWriteTargetTypes = new ConcurrentHashMap<>();
+
+		List<Object> toRegister = new ArrayList<Object>();
+
+		// Add user provided converters to make sure they can override the defaults
+		toRegister.addAll(converters);
+		toRegister.addAll(JodaTimeConverters.getConvertersToRegister());
+		toRegister.addAll(Jsr310Converters.getConvertersToRegister());
+		toRegister.addAll(ThreeTenBackPortConverters.getConvertersToRegister());
+		toRegister.addAll(storeConversions.getStoreConverters());
```
mp911de: Store converters should be probably registered after `converters` and before `JodaTimeConverters` to retain registration order. MongoDB and Cassandra modules register their converters after `converters` and before `JodaTimeConverters`.

src/main/java/org/springframework/data/convert/CustomConversions.java
```diff
@@ -0,0 +1,489 @@
...
+		Assert.notNull(converters, "List of converters must not be null!");
+
+		this.readingPairs = new LinkedHashSet<>();
+		this.writingPairs = new LinkedHashSet<>();
+		this.customSimpleTypes = new HashSet<>();
+		this.customReadTargetTypes = new ConcurrentHashMap<>();
+		this.customWriteTargetTypes = new ConcurrentHashMap<>();
+		this.rawWriteTargetTypes = new ConcurrentHashMap<>();
+
+		List<Object> toRegister = new ArrayList<Object>();
```
gregturn: At least can say `new ArrayList<>()`

src/main/java/org/springframework/data/convert/CustomConversions.java
```diff
@@ -0,0 +1,489 @@
...
+
+		// Add user provided converters to make sure they can override the defaults
+		toRegister.addAll(converters);
+		toRegister.addAll(JodaTimeConverters.getConvertersToRegister());
```
gregturn: +1

src/main/java/org/springframework/data/convert/CustomConversions.java
```diff
@@ -0,0 +1,489 @@
...
+				added = true;
+			}
+
+			if (it instanceof GenericConverter) {
+				conversionService.addConverter(GenericConverter.class.cast(it));
+				added = true;
+			}
+
+			if (!added) {
+				throw new IllegalArgumentException(
```
gregturn: I'd favor logging each registration.  Then exception can say "couldn't find match". Also avoid tracking state of who was registered.

src/main/java/org/springframework/data/convert/CustomConversions.java
olivergierke: Is there really a benefit of calling `….stream()` on each of the `List`s? The constructor invocation doesn't need an explicit generic parameter, right.

src/main/java/org/springframework/data/convert/CustomConversions.java
olivergierke: Exposing the offending instance is a good idea. Is there added value in logging the converters registered? If so, shouldn't this be functionality `GenericConversionService`?

src/main/java/org/springframework/data/convert/CustomConversions.java
olivergierke: How so? You'd explicitly need to call `StoreConversions.of(new Object[] { collection })` to run into this problem, right?

src/main/java/org/springframework/data/mapping/model/SimpleTypeHolder.java
```diff
@@ -32,39 +32,43 @@
...
-		DEFAULTS.add(Float.class);
-		DEFAULTS.add(Double.class);
-		DEFAULTS.add(Character.class);
-		DEFAULTS.add(String.class);
-		DEFAULTS.add(Date.class);
-		DEFAULTS.add(Locale.class);
-		DEFAULTS.add(Class.class);
-		DEFAULTS.add(Enum.class);
-	}
+	private static final Set<Class<?>> DEFAULTS = new HashSet<Class<?>>() {
```
olivergierke: It's the only way you can have a static `DEFAULT` instance see below *and* get the `DEFAULTS` `HashSet` populated already, which is used in the constructor. If `DEFAULTS` is set up in a static initializer block, it's still empty when the constructor is called to create the `DEFAULT` instance.

src/main/java/org/springframework/data/convert/CustomConversions.java
olivergierke: I didn't bother to again extract the converters here into a static `List` as there should be only one production instance of `CustomConversions` in place anyway. We'd basically keep yet another `List` around and introduce a layer of indirection. That said, we can definitely extract that list, no problem.

olivergierke:
From the mixture of feedback I conclude the following steps.

1. Switch to `StoreConversions.of(…, Collection<?>)` to avoid the usage of the potentially wrong overload.
2. Create a static list for the default converters we register in `CustomConversions`.
3. Stick to `addAll(…)` for final registration as we then align the abstraction level of the registration steps: user converters, store converters, default ones.
4. As 3 indicates, fix the order of registration to register the default converters last.
5. Improve exception message on converter registration.

PR #212, opened 2017-04-25
DATACMNS-1050 - Encapsulate boundaries in value objects for Range.

mp911de:
We now encapsulate a boundary in `Range` within a `Boundary` value object. `Boundary` consists of a value and whether the value is inclusive or exclusive. Boundaries without a value are unbounded.

We introduced factory methods for `Range` and `Boundary` creation using primitives and a builder to build a Range.

```java
Range<Long> range = Range.unbounded();

Range<Long> range = Range.greaterThanOrEquals(10L).andLessThanOrEquals(20L);

Range<Double> range = Range.from(10d, 20);

Boundary<Integer> lower = Boundary.inclusive(10);
Boundary<Integer> upper = Boundary.inclusive(20);

Range<Integer> range = Range.of(lower, upper);
```

---

Related ticket: [DATACMNS-1050](https://jira.spring.io/browse/DATACMNS-1050).

### DIFF
--- a/src/main/java/org/springframework/data/domain/Range.java
+++ b/src/main/java/org/springframework/data/domain/Range.java
@@ -31,23 +31,23 @@ import org.springframework.util.Assert;
 @Value
 public class Range<T extends Comparable<T>> {
 
-	private final static Range<?> UNBOUNDED = Range.of(Boundary.unbounded(), Boundary.UNBOUNDED);
+	private final static Range<?> UNBOUNDED = Range.of(Bound.unbounded(), Bound.UNBOUNDED);
 
 	/**
 	 * The lower bound of the range.
 	 */
-	private final Boundary<T> lowerBound;
+	private final Bound<T> lowerBound;
 
 	/**
 	 * The upper bound of the range.
 	 */
-	private final Boundary<T> upperBound;
+	private final Bound<T> upperBound;
 
 	/**
 	 * Creates a new {@link Range} with the given lower and upper bound. Treats the given values as inclusive bounds. Use
 	 * {@link #Range(Comparable, Comparable, boolean, boolean)} to configure different bound behavior.
 	 * 
-	 * @see Range#of(Boundary, Boundary)
+	 * @see Range#of(Bound, Bound)
 	 * @param lowerBound can be {@literal null} in case upperBound is not {@literal null}.
 	 * @param upperBound can be {@literal null} in case lowerBound is not {@literal null}.
 	 */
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/domain/Range.java
```diff
@@ -87,11 +206,230 @@ public boolean contains(T value) {
...
+	 * @soundtrack Mohamed Ragab - Excelsior Sessions (March 2017)
+	 */
+	@Value
+	public static class Boundary<T extends Comparable<T>> {
+
+		@SuppressWarnings({ "rawtypes",
+				"unchecked" }) private static final Range.Boundary<?> UNBOUNDED = new Boundary(Optional.empty(), true);
+
+		private final Optional<T> value;
+		private final boolean inclusive;
```
mp911de: What's our wording? Inclusive or including?

olivergierke: I'd like to stay with "inclusive" as`isInclusive()` reads more natural to me than `isIncluding()`.

src/main/java/org/springframework/data/domain/Range.java
```diff
@@ -87,11 +206,230 @@ public boolean contains(T value) {
...
+	/* 
+	 * (non-Javadoc)
+	 * @see java.lang.Object#toString()
+	 */
+	@Override
+	public String toString() {
+
+		String lower = this.lowerBound.getValue().map(Object::toString).map(it -> {
+
+			if (this.lowerBound.isInclusive()) {
```
olivergierke: Ternary if? Could this be a package protected `getPrefix()`/`getSuffix()` method on `Bound`?

src/main/java/org/springframework/data/domain/Range.java
```diff
@@ -87,11 +206,230 @@ public boolean contains(T value) {
...
+	 * {@link #inclusive(Comparable) including its value} or {@link #exclusive(Comparable) its value}.
+	 * 
+	 * @author Mark Paluch
+	 * @since 2.0
+	 * @soundtrack Mohamed Ragab - Excelsior Sessions (March 2017)
+	 */
+	@Value
+	public static class Boundary<T extends Comparable<T>> {
+
+		@SuppressWarnings({ "rawtypes",
```
olivergierke: Weird formatting. Maybe use `//` after the annotation?

olivergierke:
I'd like to propose:

* `Bound` over `Boundary` (shorter and I've heard the former more often in the context of ranges than the latter)
* remove the factory methods `greater…` etc. in favour of `Range.from(Bound).to(Bound)`. That leaves the details of the bound to the `Bound` type and doesn't create the need to reflect all possible ways to set up the bounds in the factory methods.

src/main/java/org/springframework/data/domain/Range.java
```diff
@@ -87,11 +206,230 @@ public boolean contains(T value) {
...
+				"unchecked" }) private static final Range.Boundary<?> UNBOUNDED = new Boundary(Optional.empty(), true);
+
+		private final Optional<T> value;
+		private final boolean inclusive;
+
+		private Boundary(Optional<T> value, boolean inclusive) {
```
olivergierke: Isn't that one created by `@Value` already?

src/main/java/org/springframework/data/domain/Range.java
```diff
@@ -68,13 +63,127 @@ public Range(T lowerBound, T upperBound) {
...
 	public Range(T lowerBound, T upperBound, boolean lowerInclusive, boolean upperInclusive) {
 
+		this.lowerBound = lowerBound == null ? Bound.unbounded()
+				: lowerInclusive ? Bound.inclusive(lowerBound) : Bound.exclusive(lowerBound);
+
+		this.upperBound = upperBound == null ? Bound.unbounded()
+				: upperInclusive ? Bound.inclusive(upperBound) : Bound.exclusive(upperBound);
+	}
+
+	private Range(Bound<T> lowerBound, Bound<T> upperBound) {
```
olivergierke: I guess this could be an `@RequiredArgsConstructor(access = AccessLevel.PRIVATE` if we annotated the fields with `@NonNull`?

src/main/java/org/springframework/data/domain/Range.java
```diff
@@ -68,13 +63,127 @@ public Range(T lowerBound, T upperBound) {
...
+
+	/**
+	 * Creates a new {@link Integer} {@link Range} with the given lower and upper bound. Treats the given values as
+	 * including bounds. Use {@link #of(Bound, Bound)} to configure different bound behavior.
+	 * 
+	 * @param lowerBound
+	 * @param upperBound
+	 * @since 2.0
+	 */
+	public static Range<Integer> from(int lowerBoundInclusive, int upperBoundInclusive) {
```
olivergierke: Can we remove those `from(…)` methods taking two arguments as they're effectively aliasing the factory methods on `Bound`, don't they?

src/main/java/org/springframework/data/domain/Range.java
```diff
@@ -87,11 +196,216 @@ public boolean contains(T value) {
...
+
+	/**
+	 * Value object representing a boundary. A boundary can either be {@link #unbounded() unbounded},
+	 * {@link #inclusive(Comparable) including its value} or {@link #exclusive(Comparable) its value}.
+	 * 
+	 * @author Mark Paluch
+	 * @since 2.0
+	 * @soundtrack Mohamed Ragab - Excelsior Sessions (March 2017)
+	 */
+	@Value
```
olivergierke: I guess we need to add an `@RequiredArgsConstructor(access = AccessLevel.PRIVATE)` to make sure the default constructor is not public.

PR #213, opened 2017-05-02
DATACMNS-1014 Remove support for JodaTime's deprecated DateMidnight

schauder:
Removed to converters using DateMidnight.

Checked MongoDB and Cassandras for compatibility. Tests ran without a problem.


mp911de:
That's merged now.

PR #214, opened 2017-05-05
DATACMNS-1058 - Alter QueryByExampleExecutor.findOne to return Optional.

christophstrobl:
We now return `Optional<S>` for `findOne()` just as we do in `CrudRepository`.


christophstrobl:
already merged on May 11th 2017 (via 0ff53bd & 7d49697)

PR #216, opened 2017-05-15
DATACMNS-1067 - Fix not to call @AfterDomainEventPublication method from Collection of entities.

sis-yoshiday:

I've fixed below ticket, please review.
https://jira.spring.io/browse/DATACMNS-1067

### DIFF
--- a/src/main/java/org/springframework/data/repository/core/support/EventPublishingRepositoryProxyPostProcessor.java
+++ b/src/main/java/org/springframework/data/repository/core/support/EventPublishingRepositoryProxyPostProcessor.java
@@ -44,6 +44,7 @@ import org.springframework.util.ReflectionUtils;
  * 
  * @author Oliver Gierke
  * @author Christoph Strobl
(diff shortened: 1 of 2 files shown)


olivergierke:
One thing I am kind of puzzled about is the mismatch between the fix and the description in the original ticket. The ticket claims a problem with `save(Iterable)` being called. The fix however deals with the events that are potentially to be emitted. Can you clarify?

olivergierke:
I think I got it.  The essential fix referring to the ticket is that the object that the clearing method is invoked on is the aggregate root, not the parameter object. However, I am now unsure about the change to only clear the events if they have been exposed. This is a feature we can discuss, but it's not something I would want to couple with the fix for the issue described in the ticket.

sis-yoshiday:
Yes, we should call clearingMethod even if events is empty.
Tomorrow, I fix it.

olivergierke:
Thanks for clarifying. Already at it :).

PR #217, opened 2017-05-17
DATACMNS-1026 ExtensionAwareEvaluationContextProvider now returns all overloaded methods as functions

schauder:
All overloaded methods are now available in SPeL expressions.

The key of the Map used for collecting methods is now the name and the length of the argument list.

It is also actually a MultiValueMap to allow multiple overloaded methods.

Among methods with identical argument list from different sources in the same extension (extension, root object, aliases) the last one in the order in parens wins.

If there is more than one method for an application the following rules are applied:

If there is one method with exact matching types in the argument list it is used. Otherwise, an exception is thrown.

### DIFF
--- a/src/main/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProvider.java
+++ b/src/main/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProvider.java
@@ -316,11 +316,7 @@ public class ExtensionAwareEvaluationContextProvider implements EvaluationContex
 		 */
 		private Optional<MethodExecutor> getMethodExecutor(EvaluationContextExtensionAdapter adapter, String name,
 				List<TypeDescriptor> argumentTypes) {
-
-			FunctionsMap functions = adapter.getFunctions();
-
-			Optional<Function> function = functions.get(name, argumentTypes);
-			return function.map(FunctionMethodExecutor::new);
+			return adapter.getFunctions().get(name, argumentTypes).map(FunctionMethodExecutor::new);
 		}
 
 		/**
@@ -389,7 +385,7 @@ public class ExtensionAwareEvaluationContextProvider implements EvaluationContex
 
 		private final EvaluationContextExtension extension;
 
-		private final FunctionsMap functions = new FunctionsMap();
+		private final Functions functions = new Functions();
 		private final Map<String, Object> properties;
 
 		/**
@@ -435,7 +431,7 @@ public class ExtensionAwareEvaluationContextProvider implements EvaluationContex
 		 * 
 		 * @return all exposed functions.
 		 */
-		FunctionsMap getFunctions() {
+		Functions getFunctions() {
 			return this.functions;
 		}
 
--- a/src/main/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProvider.java
+++ b/src/main/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProvider.java
@@ -23,7 +23,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -60,6 +59,7 @@ import org.springframework.util.StringUtils;
  * @author Thomas Darimont
  * @author Oliver Gierke
  * @author Christoph Strobl
+ * @author Jens Schauder
  * @since 1.9
  */
 public class ExtensionAwareEvaluationContextProvider implements EvaluationContextProvider, ApplicationContextAware {
@@ -317,9 +317,10 @@ public class ExtensionAwareEvaluationContextProvider implements EvaluationContex
 		private Optional<MethodExecutor> getMethodExecutor(EvaluationContextExtensionAdapter adapter, String name,
 				List<TypeDescriptor> argumentTypes) {
 
-			return adapter.getFunctions().entrySet().stream()//
-					.filter(entry -> entry.getKey().equals(name))//
-					.findFirst().map(Entry::getValue).map(FunctionMethodExecutor::new);
+			FunctionsMap functions = adapter.getFunctions();
+
+			Optional<Function> function = functions.get(name, argumentTypes);
+			return function.map(FunctionMethodExecutor::new);
 		}
 
 		/**
@@ -388,7 +389,7 @@ public class ExtensionAwareEvaluationContextProvider implements EvaluationContex
 
 		private final EvaluationContextExtension extension;
 
-		private final Map<String, Function> functions;
+		private final FunctionsMap functions = new FunctionsMap();
 		private final Map<String, Object> properties;
 
 		/**
@@ -401,17 +402,16 @@ public class ExtensionAwareEvaluationContextProvider implements EvaluationContex
 		public EvaluationContextExtensionAdapter(EvaluationContextExtension extension,
 				EvaluationContextExtensionInformation information) {
 
-			Assert.notNull(extension, "Extenstion must not be null!");
+			Assert.notNull(extension, "Extension must not be null!");
 			Assert.notNull(information, "Extension information must not be null!");
 
 			Optional<Object> target = Optional.ofNullable(extension.getRootObject());
 			ExtensionTypeInformation extensionTypeInformation = information.getExtensionTypeInformation();
 			RootObjectInformation rootObjectInformation = information.getRootObjectInformation(target);
 
-			this.functions = new HashMap<>();
-			this.functions.putAll(extensionTypeInformation.getFunctions());
-			this.functions.putAll(rootObjectInformation.getFunctions(target));
-			this.functions.putAll(extension.getFunctions());
+			functions.addAll(extension.getFunctions());
+			functions.addAll(rootObjectInformation.getFunctions(target));
+			functions.addAll(extensionTypeInformation.getFunctions());
 
 			this.properties = new HashMap<>();
 			this.properties.putAll(extensionTypeInformation.getProperties());
@@ -435,7 +435,7 @@ public class ExtensionAwareEvaluationContextProvider implements EvaluationContex
 		 * 
 		 * @return
 		 */
-		public Map<String, Function> getFunctions() {
+		public FunctionsMap getFunctions() {
 			return this.functions;
 		}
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/springframework/data/repository/query/FunctionsMap.java
```diff
@@ -0,0 +1,122 @@
...
+
+		if (candidates.isEmpty()) {
+			return Optional.empty();
+		}
+		if (candidates.size() == 1) {
+			return Optional.of(candidates.get(0));
+		}
+
+		Optional<Function> exactMatch = candidates.stream().filter(f -> f.supportsExact(argumentTypes)).findFirst();
+		if (exactMatch.isPresent()) {
```
olivergierke: `return  exactMatch.orElseThrow(…)`?

src/main/java/org/springframework/data/repository/query/FunctionsMap.java
```diff
@@ -0,0 +1,122 @@
...
+			return exactMatch;
+		} else {
+			throw new IllegalStateException("There are multiple matching methods.");
+		}
+	}
+
+	@Data
```
olivergierke: Could this be `@Value` instead?

src/main/java/org/springframework/data/repository/query/spi/Function.java
```diff
@@ -115,4 +117,49 @@ public boolean supports(List<TypeDescriptor> argumentTypes) {
...
+			if (parameterTypes[i] != argumentTypes.get(i).getType()) {
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	/**
+	 * Checks wither this {@code Function} has the same signature as another {@code Function}.
```
olivergierke: "whether"

src/main/java/org/springframework/data/util/MultiValueMapCollector.java
```diff
@@ -0,0 +1,107 @@
...
+	private final Function<T, K> keyFunction;
+	private final Function<T, V> valueFunction;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param keyFunction {@link Function} to create a key from an element of the {@link java.util.stream.Stream}
+	 * @param valueFunction {@link Function} to create a value from an element of the {@link java.util.stream.Stream}
+	 */
+	public MultiValueMapCollector(Function<T, K> keyFunction, Function<T, V> valueFunction) {
```
olivergierke: `@RequiredArgsConstructor`? I guess a static factory method in `StreamUtils` would align nicely with the ones we have in place for unmodifiable lists and sets.

src/main/java/org/springframework/data/util/MultiValueMapCollector.java
```diff
@@ -0,0 +1,107 @@
...
+
+
+	/*
+	 * (non-Javadoc)
+	 * @see java.util.stream.Collector#combiner()
+	 */
+	@Override
+	public BinaryOperator<MultiValueMap<K, V>> combiner() {
+
+		return (map1,map2) -> {
```
olivergierke: Space after comma.

src/main/java/org/springframework/data/util/MultiValueMapCollector.java
```diff
@@ -0,0 +1,107 @@
...
+
+			for (K key : map2.keySet()) {
+				map1.addAll(key, map2.get(key));
+			}
+			return map1;
```
olivergierke: Empty line before return statement.

src/main/java/org/springframework/data/util/MultiValueMapCollector.java
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+
+	/*
+	 * (non-Javadoc)
+	 * @see java.util.stream.Collector#finisher()
+	 */
+	@Override
+	public Function<MultiValueMap<K, V>, MultiValueMap<K, V>> finisher() {
+		return it -> it;
```
olivergierke: `Function.identity()`?

src/main/java/org/springframework/data/repository/query/FunctionsMap.java
```diff
@@ -0,0 +1,122 @@
...
+				functions.add(k, f);
+			}
+		});
+	}
+
+	void addAll(MultiValueMap<NameAndArgumentCount, Function> newFunctions) {
+
+		newFunctions.forEach((k, list) -> {
+			List<Function> currentElements = get(k);
+			list.forEach(f -> {
```
olivergierke: `stream().filter(…).forEach(…)`?

src/main/java/org/springframework/data/repository/query/FunctionsMap.java
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * {@link MultiValueMap} like data structure to keep lists of
+ * {@link org.springframework.data.repository.query.spi.Function}s indexed by name and argument list length, where the
+ * value lists are actually unique with respect to the signature.
+ *
+ * @author Jens Schauder
+ * @since 2.0
+ */
+class FunctionsMap {
```
olivergierke: What do you think of a plain `Functions`?

src/main/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProvider.java
```diff
@@ -424,25 +424,25 @@ public EvaluationContextExtensionAdapter(EvaluationContextExtension extension,
...
 		}
 
 		/**
 		 * Returns all functions exposed.
 		 * 
-		 * @return
+		 * @return all exposed functions.
 		 */
-		public Map<String, Function> getFunctions() {
+		FunctionsMap getFunctions() {
```
olivergierke: Wonder if it makes sense to extract an interface here so that clients can't modify the instance anymore, i.e. return something that doesn't expose the `addAll(…)` methods anymore.

src/main/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProvider.java
```diff
@@ -307,19 +307,20 @@ public void write(EvaluationContext context, Object target, String name, Object
...
 		private Optional<MethodExecutor> getMethodExecutor(EvaluationContextExtensionAdapter adapter, String name,
 				List<TypeDescriptor> argumentTypes) {
 
-			return adapter.getFunctions().entrySet().stream()//
-					.filter(entry -> entry.getKey().equals(name))//
-					.findFirst().map(Entry::getValue).map(FunctionMethodExecutor::new);
+			FunctionsMap functions = adapter.getFunctions();
+
+			Optional<Function> function = functions.get(name, argumentTypes);
+			return function.map(FunctionMethodExecutor::new);
```
olivergierke: Inline `function`?

src/main/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProvider.java
```diff
@@ -307,19 +307,20 @@ public void write(EvaluationContext context, Object target, String name, Object
...
 		 */
 		private Optional<MethodExecutor> getMethodExecutor(EvaluationContextExtensionAdapter adapter, String name,
 				List<TypeDescriptor> argumentTypes) {
 
-			return adapter.getFunctions().entrySet().stream()//
-					.filter(entry -> entry.getKey().equals(name))//
-					.findFirst().map(Entry::getValue).map(FunctionMethodExecutor::new);
+			FunctionsMap functions = adapter.getFunctions();
+
+			Optional<Function> function = functions.get(name, argumentTypes);
```
olivergierke: Inline `functions`?

src/test/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProviderUnitTests.java
```diff
@@ -219,6 +222,87 @@ public CustomExtensionRootObject1 getRootObject() {
 		assertThat(counter.get()).isEqualTo(2);
 	}
 
+	@Test // DATACMNS-1026
+	public void overloadedMethodsGetResolved() throws Exception {
+
+		provider = new ExtensionAwareEvaluationContextProvider(
```
olivergierke: Shall we extract that setup step? It's repeated in every test below.

src/main/java/org/springframework/data/repository/query/FunctionsMap.java
schauder: I thought about that, but then we return a `Function` instead of `Optional<Function>` and we end up sooner or later either rewrapping it in an Optional or doing null checks. 

What we really need is 
[`ifPresentOrElse`](http://download.java.net/java/jdk9/docs/api/java/util/Optional.html#ifPresentOrElse-java.util.function.Consumer-java.lang.Runnable-)  but that is java9

I just removed the else to make it a little lighter on the eye.

olivergierke: I see. That makes sense. Maybe inverting the check here is an option, too. `if (!exactMatch.isPresent()) { throw … }`.

I also wonder: we don't really detect multiple exact matches here, do we? It just barks if we don't find any exact match, right?

src/test/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProviderUnitTests.java
schauder: This is just in a private class anyway. A separate interface sounds like overkill to me in this case.

src/main/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProvider.java
schauder: This is just in a private class anyway. A separate interface sounds like overkill to me in this case.

src/main/java/org/springframework/data/repository/query/FunctionsMap.java
schauder: Correct: multiple exact matches are not possible at this point because we drop those when we add elements to `Functions` this ensures the behavior of overwriting of methods from different sources (extension, root object, and aliases).

But we only check for an exact match if we have more than one match (candidates). So we throw the exception if we have multiple candidates and can't disambiguate between them.

We could improve the message to: 
> There are multiple matching methods for arguments of type <argumentTypeList>, but no exact match. Make sure to provide only one matching overload or one with exactly those types.

What do you think?

olivergierke: But that's not what the code does is it? If we have multiple candidates *and* have multiple ones that potentially match exactly, we just pick the first one. Is that intended? I thought the idea was to also reject that case, wasn't it?

schauder: There can't be multiple exact matches. We don't add them to the Map contained in functions in the first place. see the `addAll` methods.

olivergierke: Ah, now that makes sense. Thanks!

PR #221, opened 2017-05-24
DATACMNS-1074 - Cache MethodHandles of default methods after lookup.

mp911de:
We now cache method handles after lookup to reuse them and prevent subsequent lookups.

MethodHandle lookup is expensive, in particular, the lookup uses exceptions as control flow [1] during speculative lookup [2].

[1] http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/5b86f66575b7/src/share/classes/java/lang/invoke/MemberName.java#l978
[2] http://mail.openjdk.java.net/pipermail/core-libs-dev/2016-August/042770.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,6 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
+	<version>2.0.0.DATACMNS-1074-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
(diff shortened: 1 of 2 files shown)


mp911de:
Thanks for having a look. The code is merge-friendly for Ingalls and Hopper. Regarding your questions:

* Using `ConcurrentReferenceMap` makes sense if we assume instances of `DefaultMethodInvokingMethodInterceptor` are going to be held `static`. I derived my assumption of the current usage pattern in which a `DefaultMethodInvokingMethodInterceptor` instance is bound to the invoker chain of a repo proxy. Have no strong opinion here.
* Using `computeIfAbsent` makes the code nicer on the usage side but we need to wrap reflection operation exceptions. Additionally, we always pay with lambda instance creation even if the element is available from the cache.

PR #222, opened 2017-05-31
DATACMNS-102 - Allow Repositories to be composed of an arbitrary number of implementation classes

mp911de:
We now use `RepositoryComposition` as backing implementation for repository method calls to implementations. `RepositoryComposition` are subject to `MethodLookup` and argument converter customization to select the target method that should be invoked. A composition consists of one or more fragments that contribute their API backed by an implementation.

During configuration, we scan for repository fragments. Fragment implementation candidates derive from the repository interface declaration, specifically the declared interfaces and their order. We scan the class path during fragment scan for each interface and add discovered fragments to the repository composition. The name of the implementation is derived from the simple name of the interface and the implementation suffix. Qualified fragments are top-level interface declarations that are not annotated with `@NoRepositoryBean`. Inherited interfaces are not considered as fragment candidates.

```java
interface ComposedRepository extends Repository<Person, Long>, 
                                                    WriteRepository<Person>, 
                                                    ReadRepository<Person>, 
                                                    QueryByExampleExecutor<Person> {}

// fragment candidate
interface WriteRepository<T> extends BaseRepository {}
// fragment implementation
class WriteRepositoryImpl<T> extends BaseRepositoryImpl implements WriteRepository<T> {}

// transient dependency of ComposedRepository, not a fragment candidate
interface BaseRepository {}
// ignored
class BaseRepositoryImpl implements BaseRepository {}

// fragment candidate
interface ReadRepository<T> extends BaseRepository {}
// fragment implementation
class ReadRepositoryImpl<T> implements ReadRepository<T> {}

// excluded fragment (can be added by RepositoryFactorySupport)
@NoRepositoryBean
interface QueryByExampleExecutor<Person> {}

```
We create a `RepositoryComposition` from the discovered fragments in the order of interface declaration in the repository interface and supply the composition to the actual repository creation.

---

Related ticket: [DATACMNS-102](https://jira.spring.io/browse/DATACMNS-102).

### DIFF
--- a/src/main/java/org/springframework/data/repository/config/RepositoryConfiguration.java
+++ b/src/main/java/org/springframework/data/repository/config/RepositoryConfiguration.java
@@ -19,6 +19,7 @@ import java.util.Optional;
 
 import org.springframework.core.type.filter.TypeFilter;
 import org.springframework.data.repository.query.QueryLookupStrategy;
+import org.springframework.data.util.Streamable;
 
 /**
  * Configuration information for a single repository instance.
@@ -33,7 +34,7 @@ public interface RepositoryConfiguration<T extends RepositoryConfigurationSource
 	 *
 	 * @return
 	 */
-	Iterable<String> getBasePackages();
+	Streamable<String> getBasePackages();
 
 	/**
 	 * Returns the interface name of the repository.
@@ -117,5 +118,5 @@ public interface RepositoryConfiguration<T extends RepositoryConfigurationSource
 	 *
 	 * @return
 	 */
-	Iterable<TypeFilter> getExcludeFilters();
+	Streamable<TypeFilter> getExcludeFilters();
 }
--- a/src/main/java/org/springframework/data/repository/config/RepositoryConfiguration.java
+++ b/src/main/java/org/springframework/data/repository/config/RepositoryConfiguration.java
@@ -22,28 +22,28 @@ import org.springframework.data.repository.query.QueryLookupStrategy;
 
 /**
  * Configuration information for a single repository instance.
- * 
+ *
  * @author Oliver Gierke
  */
 public interface RepositoryConfiguration<T extends RepositoryConfigurationSource> {
 
 	/**
 	 * Returns the base packages that the repository was scanned under.
-	 * 
+	 *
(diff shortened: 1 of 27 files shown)


schauder:
Just noted that https://jira.spring.io/browse/DATACMNS-764 is still in review. I guess that means merge conflicts. :-/

schauder:
What is the reason for excluding super interfaces as fragment candidates? I guess we would have to define an order which holds even when it appears multiple times in the interface hierarchy, but I currently don't see a problem in doing so. Do basically a breadth first search on the interface inheritance hierarchy would put implementation fragments of inherited interfaces after those mentioned directly, giving some possible useful way of overriding implementations.

mp911de:
We exclude them to lower the complexity for now. We don't want to deal with that case if an interface is implemented by multiple implementations if implementations are pulled in multiple times.

src/main/java/org/springframework/data/repository/core/support/RepositoryFactorySupport.java
```diff
@@ -183,53 +220,92 @@ public void addRepositoryProxyPostProcessor(RepositoryProxyPostProcessor process
 		this.postProcessors.add(processor);
 	}
 
+	public void setRepositoryComposition(RepositoryComposition composition) {
+
+		Assert.notNull(composition, "RepositoryComposition must not be null!");
+		this.composition = Optional.of(composition);
+	}
+
+	protected RepositoryComposition getRepositoryCompositition(RepositoryMetadata metadata) {
```
olivergierke: `Compositition` -> `Composition`

src/main/java/org/springframework/data/repository/config/RepositoryConfiguration.java
```diff
@@ -22,94 +22,99 @@
...
 	 * Returns the interface name of the repository.
-	 * 
+	 *
 	 * @return
 	 */
 	String getRepositoryInterface();
 
 	/**
 	 * Returns the key to resolve a {@link QueryLookupStrategy} from eventually.
-	 *
```
mp911de: Done.

PR #223, opened 2017-05-31
DATACMNS-991 Update DefaultMethodInvokingMethodInterceptor to Java 9 compatible method handle invocation

schauder:
The two code paths are there and they make access violations go away in the example projects.

I don't see how we can get useful tests into this at this stage. There might be some testing possible, once DATACMNS-1033 is finished, but there is still a lot of work to do for that.


mp911de:

PR #226, opened 2017-06-09
DATACMNS-1063 - Accept Publisher in findById(…) and existsById(…)

mp911de:
We now accept `Publisher<T>` instead of `Mono<T>` in `findById(…)` and `existsById(…)`. Users of a ReactiveStreams-based framework are no longer required to perform Publisher to Mono-adoption themselves but can pass a `Publisher` directly. Both methods use the first emitted value to issue their queries. Additional values are not consumed from the stream.

---

Related ticket: [DATACMNS-1063](https://jira.spring.io/browse/DATACMNS-1063).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,6 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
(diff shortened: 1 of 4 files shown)


sdeleuze:
As previously discussed, :+1: for that change.

PR #228, opened 2017-06-16
DATACMNS-1091 Sporadic test failures in Tests for JSR-310 conversions

schauder:
We now use specified date formats for converting various date types to Strings and comparing them.

Before, in some cases toString() was used which strips trailing zeros, causing sporadic failures.

JDK9 changed the behavior of Date related toString methods, making tests fail all the time.
Therefore this fix is required for executing tests with JDK9.


mp911de:

PR #232, opened 2017-07-19
DATACMNS-1114 - Introduced usage of nullable annotations for API validation.

olivergierke:
Marked all packages with Spring Frameworks @NonNullApi. Added Spring's @Nullable to methods, parameters and fields that take or produce null values. Adapted using code to make sure the IDE can evaluate the null flow properly. Fixed Javadoc in places where an invalid null handling policy was advertised. Strengthened null requirements for types that expose null-instances.

Removed null handling from converters for JodaTime and ThreeTenBP. Introduced factory methods Page.empty() and Page.empty(Pageable). Introduced default methods getRequiredGetter(), …Setter() and …Field() on PersistentProperty to allow non-nullable lookups of members. The same for TypeInformation.getRequiredActualType(), …SuperTypeInformation().

Tweaked PersistentPropertyCreator.addPropertiesForRemainingDescriptors() to filter unsuitable PropertyDescriptors before actually trying to create a Property instance from them as the new stronger nullability requirements would cause exceptions downstream.

Lazy.get() now expects a non-null return value. Clients being able to cope with null need to call ….getNullable().


src/main/java/org/springframework/data/repository/core/NamedQueries.java
```diff
@@ -37,5 +39,6 @@
 	 * @param queryName
 	 * @return
 	 */
+	@Nullable
 	String getQuery(String queryName);
```
mp911de: Would it make sense to enforce non-null here because `NamedQueries` exposes a `hasQuery(…)` method?

src/main/java/org/springframework/data/repository/query/ExtensionAwareEvaluationContextProvider.java
```diff
@@ -66,7 +68,7 @@
 
 	private final Map<Class<?>, EvaluationContextExtensionInformation> extensionInformationCache = new HashMap<>();
 
-	private List<? extends EvaluationContextExtension> extensions;
+	private @Nullable List<? extends EvaluationContextExtension> extensions;
```
mp911de: Wouldn't it rather make sense to use `Optional<List<…>>` here?

src/main/java/org/springframework/data/domain/ExampleMatcher.java
```diff
@@ -394,8 +395,8 @@ public boolean isAnyMatching() {
 	@EqualsAndHashCode
 	public static class GenericPropertyMatcher {
 
-		StringMatcher stringMatcher = null;
-		Boolean ignoreCase = null;
+		@Nullable StringMatcher stringMatcher = null;
+		@Nullable Boolean ignoreCase = null;
```
mp911de: Below in ` withMatcher(String propertyPath, GenericPropertyMatcher genericPropertyMatcher)`, the `genericPropertyMatcher.valueTransformer != null` check seems no longer required

src/main/java/org/springframework/data/util/GenericArrayTypeInformation.java
```diff
@@ -60,6 +62,7 @@ protected GenericArrayTypeInformation(GenericArrayType type, TypeDiscoverer<?> p
 	 * @see org.springframework.data.util.TypeDiscoverer#doGetComponentType()
 	 */
 	@Override
+	@Nullable
```
mp911de: Is `@Nullable` required here?

src/main/java/org/springframework/data/util/ParameterizedTypeInformation.java
```diff
@@ -39,7 +41,7 @@
 class ParameterizedTypeInformation<T> extends ParentTypeAwareTypeInformation<T> {
 
 	private final ParameterizedType type;
-	private Boolean resolved;
+	private @javax.annotation.Nullable Boolean resolved;
```
mp911de: Is the qualified type required? Maybe even a candidate for `Lazy<Boolean>`

src/main/java/org/springframework/data/util/ReflectionUtils.java
```diff
@@ -189,13 +193,32 @@ public static Field findField(Class<?> type, DescribedFieldFilter filter, boolea
...
 
 	/**
+	 * Finds the field of the given name on the given type.
+	 * 
+	 * @param type must not be {@literal null}.
+	 * @param name must not be {@literal null} or empty.
+	 * @return
+	 * @throws IllegalArgumentException in case the field can't be found.
+	 */
+	public static Field findRequiredField(Class<?> type, String name) {
```
mp911de: how about `getField` or `getRequiredField`?

src/main/java/org/springframework/data/util/ReflectionUtils.java
```diff
@@ -219,6 +242,32 @@ public static void setField(Field field, Object target, Object value) {
...
 	/**
+	 * Returns the method with the given name of the given class and parameter types.
+	 * 
+	 * @param type must not be {@literal null}.
+	 * @param name must not be {@literal null}.
+	 * @param parameterTypes must not be {@literal null}.
+	 * @return
+	 * @throws IllegalArgumentException in case the method cannot be resolved.
+	 */
+	public static Method findRequiredMethod(Class<?> type, String name, Class<?>... parameterTypes) {
```
mp911de: How about `getMethod` or `getRequiredMethod`?

src/main/java/org/springframework/data/util/Version.java
```diff
@@ -139,7 +141,7 @@ public boolean isLessThanOrEqualTo(Version version) {
 	 * (non-Javadoc)
 	 * @see java.lang.Comparable#compareTo(java.lang.Object)
 	 */
-	public int compareTo(Version that) {
+	public int compareTo(@Nullable Version that) {
```
mp911de: `Comparable` does not allow `null` values.

src/main/java/org/springframework/data/geo/GeoResults.java
```diff
@@ -53,9 +57,9 @@ public GeoResults(List<? extends GeoResult<T>> results) {
 	 * from the distance values of the given {@link GeoResult}s.
 	 * 
 	 * @param results must not be {@literal null}.
-	 * @param metric must not be {@literal null}.
+	 * @param metric can be {@literal null}.
 	 */
-	public GeoResults(List<? extends GeoResult<T>> results, Metric metric) {
+	public GeoResults(List<? extends GeoResult<T>> results, @Nullable Metric metric) {
```
olivergierke: Good point, I'll change that.

src/main/java/org/springframework/data/mapping/PropertyPath.java
```diff
@@ -252,6 +281,11 @@ public static PropertyPath from(String source, TypeInformation<?> type) {
 			}
 		}
 
+		if (result == null) {
+			throw new IllegalStateException(
```
olivergierke: With `Assert.state(result == null, …)` the IDE doesn't actually consider the variable tested as non-null.

jhoeller: Is that specific to Eclipse? In IntelliJ, I'm using `Assert.state(result != null, ...)` all over the place... checking for not-null, of course, but I guess that was just a typo :-)

src/main/java/org/springframework/data/support/IsNewStrategyFactorySupport.java
```diff
@@ -55,5 +56,6 @@ public final IsNewStrategy getIsNewStrategy(Class<?> type) {
 	 * @param type will never be {@literal null}.
 	 * @return the {@link IsNewStrategy} to be used for the given type or {@literal null} in case none can be resolved.
 	 */
+	@Nullable
 	protected abstract IsNewStrategy doGetIsNewStrategy(Class<?> type);
```
olivergierke: The method is used internally only anyway as we throw an exception in case of `null` being returned from the template method. I wanted to avoid introducing a null object for a type which — through all public API — will never be null in the first place.

src/main/java/org/springframework/data/util/Lazy.java
```diff
@@ -56,12 +58,37 @@
...
+		return value;
+	}
+
+	/**
+	 * Returns the value of the lazy evaluation.
+	 * 
+	 * @return
+	 */
+	@Nullable
+	public T getNullable() {
```
olivergierke: Good one, I'll move to that.

src/main/java/org/springframework/data/mapping/PropertyPath.java
olivergierke: Eclipse doesn't recognize that as it basically requires Spring specific knowledge in the IDE which I don't think we can/should rely on.

src/main/java/org/springframework/data/repository/config/AnnotationRepositoryConfigurationSource.java
```diff
@@ -74,18 +78,25 @@
...
+		super(environment, ConfigurationUtils.getRequiredClassLoader(resourceLoader), registry);
 
 		Assert.notNull(metadata, "Metadata must not be null!");
 		Assert.notNull(annotation, "Annotation must not be null!");
 		Assert.notNull(resourceLoader, "ResourceLoader must not be null!");
 
-		this.attributes = new AnnotationAttributes(metadata.getAnnotationAttributes(annotation.getName()));
+		Map<String, Object> annotationAttributes = metadata.getAnnotationAttributes(annotation.getName());
+
+		if (annotationAttributes == null) {
```
olivergierke: See above.

src/main/java/org/springframework/data/util/ReflectionUtils.java
olivergierke: The corresponding methods in Spring Framework are called `findField(…)` which is why I used this pattern.

src/main/java/org/springframework/data/util/ReflectionUtils.java
olivergierke: See above.

olivergierke:
Feedback addressed.

PR #233, opened 2017-07-26
DATACMNS-1126 - Add Kotlin constructor support.

mp911de:
We now discover Kotlin constructors and apply parameter defaulting to allow construction of immutable value objects. Constructor discovery uses primary constructors by default and considers `@PersistenceConstructor` annotations.

`ClassGeneratingEntityInstantiator` can instantiate Kotlin classes with default parameters by resolving the synthetic constructor. Null values translate to parameter defaults.

```kotlin
class Person(val firstname: String = "Walter") {
}

class Address(val street: String, val city: String) {

	@PersistenceConstructor
	constructor(street: String = "Unknown", city: String = "Unknown", country: String) : this(street, city)
}
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.0.0.BUILD-SNAPSHOT</version>
+	<version>2.0.0.DATACMNS-1126-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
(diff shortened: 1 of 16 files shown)


mp911de:
I'd turn `PreferredConstructorDiscoverer` into an interface with static factory methods and `enum`s for the particular implementations (Java and Kotlin).

An `EntityInstantiator` for Kotlin classes is mostly the same as an instantiator for Java classes. Only Kotlin constructors with default arguments require a more specific instantiation. We could create a subclass `ClassGeneratingKotlinEntityInstantiator` extending `ClassGeneratingEntityInstantiator` and differentiate in `EntityInstantiators` which one to use.

PR #238, opened 2017-08-21
DATACMNS-1140 - Open up Query by Example to allow extension in module implementations.

christophstrobl:
We turned `Example` into an interface with default implementation created via the static of methods. This allows store implementations to create their very own flavor of `Example` allowing eg. non typed ones.


mp911de:
That's merged and polished.

PR #241, opened 2017-09-12
DATACMNS-1157 - Enforce nullable/non-null API constraints on repository query methods

mp911de:
We now validate query method invocations to check method parameters whether they accept `null` values and reject execution if `null` values are not accepted. We derive nullability rules from repository interfaces declared using Kotlin and Spring's `@NonNullApi`/`@Nullable` annotations.

We also check whether a query method can return `null`. If a query method returns `null` which isn't supposed to do so, then we throw `EmptyResultDataAccessException` to prevent `null` return values.

```java
interface UserRepository extends Repository<User, String> {

  List<User> findByLastname(@Nullable String firstname);

  @Nullable
  User findByFirstnameAndLastname(String firstname, String lastname);
}
```

```kotlin
interface UserRepository : Repository<User, String> {

	fun findByLastname(username: String?): List<User>

	fun findByFirstnameAndLastname(firstname: String, lastname: String): User?
}
```

---

Related tickets: [DATACMNS-1154](https://jira.spring.io/browse/DATACMNS-1154), [DATACMNS-1157](https://jira.spring.io/browse/DATACMNS-1157).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.0.0.BUILD-SNAPSHOT</version>
+	<version>2.0.0.DATACMNS-1154-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
--- a/src/main/asciidoc/repositories.adoc
+++ b/src/main/asciidoc/repositories.adoc
@@ -852,14 +852,14 @@ class AnAggregateRoot {
         // … return events you want to get published here
     }
 
-    @AfterDomainEventsPublication <2>
+    @AfterDomainEventPublication <2>
     void callbackMethod() {
        // … potentially clean up domain events list
     }
 }
 ----
 <1> The method using `@DomainEvents` can either return a single event instance or a collection of events. It must not take any arguments.
-<2> After all events have been published, a method annotated with `@AfterDomainEventsPublication`. It e.g. can be used to potentially clean the list of events to be published.
+<2> After all events have been published, a method annotated with `@AfterDomainEventPublication`. It e.g. can be used to potentially clean the list of events to be published.
 ====
 
 The methods will be called every time one of a Spring Data repository's `save(…)` methods is called.
--- a/src/main/asciidoc/repositories.adoc
+++ b/src/main/asciidoc/repositories.adoc
@@ -864,6 +864,70 @@ class AnAggregateRoot {
 
 The methods will be called every time one of a Spring Data repository's `save(…)` methods is called.
 
+[[core.nullability-validation]]
+== Null-safety
+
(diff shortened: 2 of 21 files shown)


olivergierke:
Generally looks good to me. Shall we rather throw `EmptyResultDataAccessException`? It seems to be a more specific match for what we're trying to achieve.

olivergierke:
@jhoeller mentioned, that Spring 5 also already considers those annotations on injection points, parameter injection in MVC etc. So we're not even setting precedence here.

olivergierke:
I wouldn't mind @sdeleuze having a quick glance as it touches out Kotlin story as well.

sdeleuze:
Good timing for discussing that. First be aware that I am doing currently the latest changes in Spring Framework and Reactor null-safety annotations. Nothing super disruptive for you but my goal is to support generic type argument null-safety (critical for Reactor use case for example), allow fine grain null-safety with a more focused scope than the package and also to avoid fields to be in `@NonNullApi` scope  (most are `private or protected` so not part of the public API) while still allowing field nullability (to make Eclipse happy mainly). See [this commit](https://github.com/sdeleuze/spring-framework/commit/SPR-15756), and I am currently working on adding `@NonNull` annotations at field and generic type argument level.

I agree with the principle of the current PR and think that's a very good idea for Spring Data users. My main question is about `NullableUtils`, did you implement that logic yourself instead of leveraging Kotlin reflection capabilities because you want support that for Java as well in addition to Kotlin?

mp911de:
I added null-safety with query methods to the reference docs and incorporated the first round of feedback.

olivergierke:
@sdeleuze — Yes, our intention is to allow users that use nullable annotations with their Java based repository interfaces to also get exceptions thrown in case there's a mismatch between the expectations described by the annotations (i.e. `@NonNullApi` reaching into the repository) but a query method actually returning `null` from an execution. 

The user would then be left with three choices: alter the query to make sure it never produces `null`, add `@Nullable` to indicate nullability or declare the return type a as JDK 8 `Optional`, Vavr `Option` or alternatives. In the case of Kotlin, we inspect the return type declaration to basically achieve the same: runtime validation of the contract defined.

olivergierke:
@mp911de — Regarding the timing of the feature I'm wondering if in the current implementation the interceptor would automatically kick in for CRUD methods, as they're already using the nullable annotations. If that wasn't the case, and the nullability checks are only applied in case the actual repository interface is under `@NonNullApi` governance – something that a user would have to opt-in to deliberately – I think we could sneak it into Kay GA. We could then think about whether we'd generally want to activate the interception for everything in `CrudRepository` – even without explicit activation by the user – for Lovelace.

mp911de:
The activation is entirely driven by the package of the actual repository interface or when the repository interface is written in Kotlin and `kotlin-reflect` is available on the class path. Intermediate interfaces (even when they carry nullability annotations or are located in `@NonNullApi` packages) do not control the activation of `null` checks for the whole interface.

I think we will need to the adapt the tests in our modules.

PR #244, opened 2017-09-22
DATACMNS-1170 - Fallback to default constructor discovery for Kotlin classes without a primary constructor

mp911de:
We now fall back to default preferred constructor discovery if a Kotlin class has no primary constructor and no preferred constructor is resolved. Previously, primary constructor resolution yielded no result (returned `null`) which caused the subsequent Java constructor lookup to fail.

---

Related ticket: [DATACMNS-1170](https://jira.spring.io/browse/DATACMNS-1170).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.0.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


sdeleuze:
Not sure if you are interested but I am about committing that on Spring Framework side, maybe you wuld be intersted to leverage this updated `BeanUtils.findPrimaryConstructor`. https://github.com/sdeleuze/spring-framework/commit/8904847576e929c24c150797a4987a33ae467bd0

PR #246, opened 2017-09-25
Add PropertyPathMapper.

alesj:
Adding PR for this issue:
* https://jira.spring.io/browse/DATACMNS-1176

I'm open for suggestions on how to dynamically validate part's path,
and not be tied to class's hierarchy.


olivergierke:
Let's get to some agreement on the ticket first before we start proposing any code changes. I've left an explanation what parts of the codebase are responsible for what on the ticket. Let's work from there to see we get your use case working. Changing APIs in very fundamental parts of Spring Data Commons usually isn't the first option.

PR #253, opened 2017-10-06
DATACMNS-1190 - Added section on how to enforce nullability constraints on repositories

olivergierke:


src/main/asciidoc/repositories.adoc
```diff
@@ -214,6 +214,82 @@ In this first step you defined a common base interface for all your domain repos
...
+* `com.google.common.base.Optional`
+* `scala.Option`
+* `io.vavr.control.Option`
+* `javaslang.control.Option` (deprecated as Javaslang is deprecated)
+
+Alternatively query methods can choose not to use a wrapper type at all.
+The absence of a query result will then be indicated by returning `null`.
+Repository methods returning collection like values will never return null but an empty value.
+
+[[repositories.nullability.annotations]]
```
mp911de: This section duplicates [the section on Null-safety](https://github.com/spring-projects/spring-data-commons/blame/c9e0803beedb7284cc6a51a4a67aa20dd1cb6898/src/main/asciidoc/repositories.adoc#L869) that is currently in place. It would make sense to move `core.nullability-validation` in here.

src/main/asciidoc/repositories.adoc
```diff
@@ -214,6 +214,82 @@ In this first step you defined a common base interface for all your domain repos
...
+Besides that, Spring Data supports to return other wrapper types on query methods:
+
+* `com.google.common.base.Optional`
+* `scala.Option`
+* `io.vavr.control.Option`
+* `javaslang.control.Option` (deprecated as Javaslang is deprecated)
+
+Alternatively query methods can choose not to use a wrapper type at all.
+The absence of a query result will then be indicated by returning `null`.
+Repository methods returning collection like values will never return null but an empty value.
```
mp911de: How about `Repository methods returning collections, wrappers, and streams are guaranteed to never return null but rather the corresponding empty representation.`?

olivergierke: Maybe we should link to the appendix table listing all supported return types as they also list the alternative collection types we support that might be handy to know about but don't need any discussion here.

src/main/asciidoc/repositories.adoc
olivergierke: After revisiting the other version of the docs, I'd actually like to keep this one here and augment it with the following aspects currently present in the other version:

- Nullability annotations should list the fully-qualified name at least once, so that it's clear what to import. I think we can (should?) leave out the remark with JSR-305 needed for meta-annotation usage. That's taking it a bit too far I think.
- The `kotlin-reflect` library needed to get the runtime checks enabled.

src/main/asciidoc/repositories.adoc
```diff
@@ -214,6 +214,82 @@ In this first step you defined a common base interface for all your domain repos
...
+<1> Will throw an `EmptyResultDataAccessException` in case the query executed does not produce a result. Will throw an `IllegalArgumentException` in case the `emailAddress` handed to the method is `null`.
+<2> Will return `null` in case the query executed does not produce a result. Also accepts `null` as value for `emailAddress`.
+<3> Will return `Optional.empty()` in case the query executed does not produce a result. Will throw an `IllegalArgumentException` in case the `emailAddress` handed to the method is `null`.
+====
+
+[[repositories.nullability.kotlin]]
+==== Nullability in Kotlin-based repositories
+
+Kotlin has the definition of nullability constraints baked into the language.
+Spring Data repositories using the language mechanism to define those constraints will automatically get the same runtime checks applied:
```
olivergierke: Probably a good place to mention the need to have `kotlin-reflect` on the classpath.

PR #258, opened 2017-11-13
DATACMNS-1208 - fixed hasPersistentEntityFor() vs getPersistentEntity…

bkielczewski:
…ForClass() inconsistency

Making call to AbstractMappingContext.getPersistentEntityForClass() created an entry in persistentEntities map for every type for which shouldCreatePersistentEntity() == false. This was causing hasPersistentEntityFor() to falsely return true after such call. This was fixed by not keeping Optionals in persistentEntities as this seem to serve no purpose.


olivergierke:
Fixed with a less invasive change. See the ticket for details.

PR #259, opened 2017-11-14
DATACMNS-1210 - Fix concurrency issue in BasicPersitentEntity.

christophstrobl:
Make sure to use concurrent collection implementations for properties holding mutable state. Especially the ones making use of `computeIfAbsent()`.

Usage of `ConcurrentReferenceHashMap` allows us to move on without additional changes as it allows to store `null` values, whereas `ConcurrentHashMap` would require us to store explicit `NullValue` placeholders for such cases, potentially causing trouble in downstream store specific projects.

JMH Benchmarks (for the mapping layer in Spring Data MongoDB) comparing `ConcurrentReferenceHashMap` and `ConcurrentHashMap` did not show any significant difference between the two.

### DIFF
--- a/src/test/java/org/springframework/data/mapping/model/BasicPersistentEntityUnitTests.java
+++ b/src/test/java/org/springframework/data/mapping/model/BasicPersistentEntityUnitTests.java
@@ -19,11 +19,14 @@ import static org.assertj.core.api.Assertions.*;
 import static org.junit.Assume.*;
 import static org.mockito.Mockito.*;
 
+import java.lang.annotation.Annotation;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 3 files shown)


src/test/java/org/springframework/data/mapping/model/BasicPersistentEntityUnitTests.java
```diff
@@ -276,6 +283,64 @@ public void getRequiredAnnotationThrowsException() {
 		assertThatThrownBy(() -> entity.getRequiredAnnotation(Document.class)).isInstanceOf(IllegalStateException.class);
 	}
 
+	@Test // DATACMNS-1210
+	public void findAnnotationShouldBeThreadSafe() throws InterruptedException {
+
+		assumeTrue("Requires Java 9",
```
olivergierke: What's the reason we have to assume Java 9 here?

christophstrobl: I can remove it - the point was that the test would pass on Java8 regardless of the changes made.

olivergierke: Okay, I'll just go ahead and remove it during the merge.

PR #261, opened 2017-11-17
Wrapper type in Pageable

StefanoGuerrini:
Is possible to change interface Pageable to support wrapper types?

Whit this modification it's possible to support null value, for example we can write this:

`if (pageable.getOffset()!=null) { query.offset(pageable.getOffset()); }`
`if (pageable.getPageSize()!=null) { query.limit(pageable.getPageSize()); }`

### DIFF
--- a/src/main/java/org/springframework/data/domain/AbstractPageRequest.java
+++ b/src/main/java/org/springframework/data/domain/AbstractPageRequest.java
@@ -29,8 +29,8 @@ public abstract class AbstractPageRequest implements Pageable, Serializable {
 
 	private static final long serialVersionUID = 1232825578694716871L;
 
-	private final int page;
-	private final int size;
+	private final Integer page;
+	private final Integer size;
 
 	/**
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@StefanoGuerrini Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=261)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=261) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@StefanoGuerrini Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=261)!

olivergierke:
Please open a ticket for ideas you have. Generally speaking, this is exactly the opposite direction of what we actually want to go: we prefer less nullability instead of more nullability. You can use `Pageable.isPaged()` to check whether pagination info was given. If that's the case, reasonable ofsets and limits will be set.

PR #264, opened 2017-12-03
DATACMNS-1211 - Provide WebFlux Pageable and Sort argument resolver

mp911de:
We now provide synchronous `HandlerMethodArgumentResolver`s (`SyncHandlerMethodArgumentResolver`) for `Pageable` and `Sort` objects to be used with WebFlux controllers. Resolution happens synchronously as we create `Pageable` and `Sort` from the query string which can be read without blocking. We support `@SortDefault`, `@SortDefaults` and `@PageableDefault` to apply defaulting along with fallback values.

```java
class MyWebFluxConfigurer implements WebFluxConfigurer {

	@Override
	public void configureArgumentResolvers(ArgumentResolverConfigurer configurer) {

		configurer.addCustomResolver(new ReactiveSortHandlerMethodArgumentResolver(),
				new ReactivePageableHandlerMethodArgumentResolver());
	}
}

class MyController {

	@GetMapping("people/paged")
	public Flux<Person> findAll(@PageableDefault(size = 100, sort="firstname") Pageable pageable) {
	}

	@GetMapping("people/sorted")
	public Flux<Person> findAll(@SortDefault(sort="firstname") Sort sort) {
	}
}
```

---

Related ticket: [DATACMNS-1211](https://jira.spring.io/browse/DATACMNS-1211).


Opalo:
What is the status of this PR?

mp911de:
We want to include the change and I think this will be part of the Moore release train.

reneschroeder0000:
Nice work. Did you think about implementing respective HATEOAS functionality as well? Currently when building the links, the queryParams won't be added.

nekperu15739:
Like title said, i would like to have support for webflux, but for RouterFunctions. right now i'm doing some develop based on PageableHandlerMethodArgumentResolver, however, is better in my opinion that this feature comes with framework.

regulskimichal:
Will this feature be included in Spring Data Moore? It's already RC3 stage... It would be nice to see it in the final version.

mp911de:
That's merged now.

cbornet:
Are there plans to autoconfigure this in Spring Boot ?
I guess the next step would be to have `SpringDataWebFluxConfiguration` and `EnableSpringDataWebfluxSupport` in spring-data then `SpringDataWebFluxAutoConfiguration` in boot ?
I can have a look if that's the way to go.

mp911de:
The issue here is that these argument resolvers are only partial components. Their WebMVC variants are used to enable general Spring Data REST functionality. On their own, they are simple utilities that provide a certain value.

We are not going to provide a Webflux-based Spring Data REST integration. That being said, can you file a Spring Boot ticket for further discussion? Involving the Boot team usually leads to a better overall integration to figure our an ideal approach.

PR #267, opened 2017-12-29
DATACMNS-563 - fix for returning correct page number in metadata when one indexed parameters are used

marceloverdijk:
This fix returns the correct page number when one index parameters are used.

So instead of:

```
"page": {
  "size": 10,
  "total_elements": 888,
  "total_pages": 89,
  "number": 0
}
```
 it will return:

```
"page": {
  "size": 10,
  "total_elements": 888,
  "total_pages": 89,
  "number": 1
}
```

for the first page.

### DIFF
--- a/src/main/java/org/springframework/data/web/PagedResourcesAssembler.java
+++ b/src/main/java/org/springframework/data/web/PagedResourcesAssembler.java
@@ -290,10 +290,11 @@ public class PagedResourcesAssembler<T> implements ResourceAssembler<Page<T>, Pa
 	 * @param page must not be {@literal null}.
 	 * @return
 	 */
-	private static <T> PageMetadata asPageMetadata(Page<T> page) {
+	private <T> PageMetadata asPageMetadata(Page<T> page) {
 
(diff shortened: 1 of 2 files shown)


olivergierke:
That's polished and merged, thanks!

marceloverdijk:
Thx! Will Boot 2.0 be based on Kay or Lovelace? If it’s the first would you consider backporting this to Kay as well?

olivergierke:
Boot 2.0 is on Kay. As indicated in the original ticket, I've back-ported the fix to Kay and Ingalls even, so that even Boot 1.5 gets the fix.

hurelhuyag:
I assumed `spring.data.web.pageable.one-indexed-parameters=true` is used in repository implementation and everywhere else should pass as is. But response page index still zero indexed.

odrotbohm:
As the name suggests, the setting is about the parameters exposed in the web request only. The indexes are then normalized into 0 based ones for use in `PageRequest` / `Page` etc.

PR #269, opened 2018-01-18
DATACMNS-1245 - Consider Kotlin metadata annotation for Kotlin class kind retrieval

mp911de:
We now inspect Kotlin's `Metadata` annotation to determine the kind of a Kotlin class.

Previously we used Kotlins internal API to introspect classes. Because this API is not public and can be changed at any time we rely on Kotlins annotations as they are supposed to not change in near future.

---

Related ticket: [DATACMNS-1245](https://jira.spring.io/browse/DATACMNS-1245).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.1.0.BUILD-SNAPSHOT</version>
+	<version>2.1.0.DATACMNS-1245-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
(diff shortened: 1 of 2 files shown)


mp911de:
The code is called a few times during initializations of `EntityInstantiator`, `MethodInvocationValidator`, `PreferredConstructorDiscoverer` and `AbstractMappingContext.shouldCreatePersistentEntityFor`. 

I omitted `static final` caches in the first place to avoid references preventing `ClassLoader` cleanup but maybe it's worth to encapsulate the code in an own class held in a `WeakReference`.

olivergierke:
It looks like we have a early return in place we realize it's not a Kotlin class in the first place so we should be fine, I think. We can approach this in case it comes up in benchmarks lazily.

olivergierke:
Upon closer inspection, even the early return requires repeated reflection lookups in `loadKotlinClass(…)`. I wonder if the general reflection lookup couldn't just keep what ends up in `create` and `typeFactory` in static fields for subsequent reuse?

mp911de:
Putting these values into static fields creates references that possibly prevent finalizing, therefore I'd rather go for `WeakReference` fields.

mp911de:
Changed the approach after receiving feedback from Jetbrains. Evaluating `@Metadata(k=…)` is the more stable approach.

olivergierke:
Wow, that looks dramatically simpler :)

PR #273, opened 2018-02-14
 DATACMNS-1259 - Fixed support for Long values in Auditables.

schauder:
Using Instant as internal data type since it's a point in time without time zone which LocalDateTime isn't.
Added necessary converters.
Fixed one JodaTime converter that used UTC to use SystemDefault like other similar converters.

In case of a conversion failure, the error message now contains the source type.


olivergierke:
I am puzzled

PR #274, opened 2018-02-17
DATACMNS-1261 - Apply Java 8 Functional types and Lambdas to simplify…

jxblum:
… parsing logic in Repository configuration.

### DIFF
--- a/src/main/java/org/springframework/data/config/ConfigurationUtils.java
+++ b/src/main/java/org/springframework/data/config/ConfigurationUtils.java
@@ -37,15 +37,13 @@ public interface ConfigurationUtils {
 	 * @return
 	 * @throws IllegalArgumentException if no {@link ResourceLoader} can be obtained from the {@link XmlReaderContext}.
 	 */
-	public static ResourceLoader getRequiredResourceLoader(XmlReaderContext context) {
+	static ResourceLoader getRequiredResourceLoader(XmlReaderContext context) {
 
 		Assert.notNull(context, "XmlReaderContext must not be null!");
 
 		ResourceLoader resourceLoader = context.getResourceLoader();
 
-		if (resourceLoader == null) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/springframework/data/config/ConfigurationUtils.java
```diff
@@ -88,16 +84,14 @@ public static ClassLoader getRequiredClassLoader(ResourceLoader resourceLoader)
...
 
 		Assert.notNull(beanDefinition, "BeanDefinition must not be null!");
 
 		String result = beanDefinition.getBeanClassName();
 
-		if (result == null) {
-			throw new IllegalArgumentException(
-					String.format("Could not obtain required bean class name from BeanDefinition!", beanDefinition));
-		}
+		Assert.notNull(result, String.format("Could not obtain required bean class name from BeanDefinition [%s]!",
```
mp911de, line 93: This change has performance impact as `String.format(…)` is invoked regardless of whether it's required to. We should either stick with the previous variant or use a message supplier. There are also occurrences of similar changes.

src/main/java/org/springframework/data/util/IterableUtils.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.lang.Nullable;
+
+/**
+ * Utility class for working with {@link Iterable} objects.
+ *
+ * @author John Blum
+ * @see java.lang.Iterable
+ * @since 2.1.0
+ */
+public abstract class IterableUtils {
```
mp911de, line 30: Not sure that Spring Data is the best place for such a utility. I'd suggest to move it to Spring Framework.

mp911de:
Closing this pull request since there was no update since quite a while. We can reopen it if we want to revive it.

PR #277, opened 2018-03-05
DATAJPA-1003 - Modifying the method that create a Projection for a interface to also create a projection for a non final Class using CGLIB proxy.

lbernardomaia:
Hi,

I´ve made some changes that will allow work with projection using Class instead of only Interface. The bug was reported for query annoted with @Query but Native queries using @Query was not working as well.

With these changes If the class is not a interface and a non final class, the createProjection method going to use CGLIB-based proxy. Otherwise will continue using the default converters.

thanks.

- [ ] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [ ] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATACMNS).

### DIFF
--- a/src/main/java/org/springframework/data/projection/ProxyProjectionFactory.java
+++ b/src/main/java/org/springframework/data/projection/ProxyProjectionFactory.java
@@ -41,6 +41,7 @@ import org.springframework.util.ConcurrentReferenceHashMap;
  *
  * @author Oliver Gierke
  * @author Christoph Strobl
+ * @author Bernardo Maia
  * @see SpelAwareProxyProjectionFactory
  * @since 1.10
  */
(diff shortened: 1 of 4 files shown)


olivergierke:
Before submitting a pull request for a given project, please make sure you have filed a ticket in that given project. A reference to a downstream project's ticket is not sufficient as changes in Spring Data Commons will affect *all* downstream Spring Data modules.

In that particular case I'd argue we shouldn't do this as for class based return types it's better to simply let the store create instances of the particular type in the first place. There's just no need to create a class if an interface is sufficient.

lbernardomaia:
Hi Oliver,

Thank you for the feedback.

Sorry, I did not realize that I should create a ticket for this project.

About the changes, I agree with your point that the interface projection should be the default choice.

Although, in some cases, we can face a migration to use Spring Data Repository in some project, for example. In this scenario if the project has a hundred of DTOs Class that are populated by resultSetMapping with information from queries. Would be much easier to let the projection create a Class using the constructor in a similar way that resultSetMapping would do, instead of migrate all existing DTOs Class to Interfaces.

Also I think that is more readable to let Projection create a Class through the constructor instead of using a constructor expression in the query declaration.

Thanks.

mp911de:
We decided to not include this change as per discussion on the [linked ticket](https://jira.spring.io/browse/DATAJPA-1003).

PR #282, opened 2018-04-06
DATACMNS-1290 - Added support for timestamp values of type long.

schauder:
The type long or Long is actually required for custom revision entities by Envers.

See also: spring-projects/spring-data-envers#122


mp911de:

PR #286, opened 2018-04-13
DATACMNS-1294 - Consider java.time types as simple ones.

mp911de:
We now consider all types in the `java.time` package as simple types to prevent deep reflective access. We are already excluding java.lang types for the same reason.

---

Related ticket: [DATACMNS-1294](https://jira.spring.io/browse/DATACMNS-1294).

### DIFF
--- a/src/main/java/org/springframework/data/mapping/model/SimpleTypeHolder.java
+++ b/src/main/java/org/springframework/data/mapping/model/SimpleTypeHolder.java
@@ -155,7 +155,7 @@ public class SimpleTypeHolder {
 			return isSimpleType;
 		}
 
-		if (type.getName().startsWith("java.lang")) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/mapping/model/SimpleTypeHolder.java
```diff
@@ -155,7 +155,7 @@ public boolean isSimpleType(Class<?> type) {
 			return isSimpleType;
 		}
 
-		if (type.getName().startsWith("java.lang")) {
+		if (type.getName().startsWith("java.lang") | type.getName().startsWith("java.time")) {
```
olivergierke: No double pipe for a reason?

mp911de: Yikes! Consider this one a typo. That's fixed now.

PR #288, opened 2018-04-19
DATACMNS-1303: Change two method names.

BruceKuiLiu:
The two methods are adding given events to the "domainEvents", so that their names should be "add***" but not "and***".


mp911de:
Please create a ticket in our [issue tracker](https://jira.spring.io/browse/DATACMNS/) so we can understand, what exactly you want to achieve.

For this change, I don't think this is going to happen as the change breaks compatibility.

BruceKuiLiu:
https://jira.spring.io/browse/DATACMNS-1303

mp911de:
Since the API is now public we cannot change the method names anymore. We keep DATACMNS-1303 open for revisiting the topic for Spring Data 3.0.

PR #292, opened 2018-05-22
DATACMNS-1322 - Add mutation support for immutable objects.

mp911de:
We now support two patterns to associate immutable objects with generated/loaded property values that are required to load or persist objects. The most common use case for this is if a database id gets generated or a version counter (optimistic locking) gets incremented, then these details must be returned to the domain object.

For mutation, we support wither methods and Kotlin's data classes leveraging the `copy` method.

_Note:_ TANSTAAFL applies for supporting mutation with immutable objects as immutability requires potentially a couple of intermediate objects that are discarded once the final object is built.

**Java Example**

```java
class Person {
	final String id;
	String name;

	ValueClass withId(String id) {
		return new ValueClass(id);
	}
}
```

**Java/Lombok Example**

```java
@Wither
class Person {
	String id;
	String name;
}
```

**Kotlin Example**

```kotlin
data class DataClassKt(val id: String, val name: String) {
}
```

---

Related ticket: [DATACMNS-1322](https://jira.spring.io/browse/DATACMNS-1322).


mp911de:
Merged for Lovelace.

PR #299, opened 2018-06-26
Add a findByIdOrNull extension to CrudRepository

sdeleuze:
In Kotlin, it is idiomatic to deal with return value that could have or not a result with nullable types since they are natively supported by the language.

This commit add a `findByIdOrNull` variant to `CrudRepository#findById` that returns `T?` instead of `Optional<T>`.


odrotbohm:
That's merged.

PR #322, opened 2018-11-02
Add PageRequest builder

Aloren:
This is a proposal PR for adding builder for `PageRequest` class.
Every time I see `PageRequest.of(0, 10)` in code I need to go into `PageRequest` to make sure that parameters were not mixed up. I think that having both parameters of the same type in constructor/static initializer is error prone. In this PR I would like to show you an example of how readable `PageRequest` could be. Please take a look. Thanks


JonasXPX (approved):

mp911de:

PR #330, opened 2019-01-11
DATACMNS-1438, DATACMNS-1461 - Improved property path access.

odrotbohm:
We now expose a dedicated `PersistentPropertyPathAccessor` and optionally take options for both the read and write access to properties. On read, clients can define to eagerly receive null in case one of the path segments is null. By default, we reject those as it indicates that there might be an issue with the backing object if the client assumes it can look up the more deeply nested path to receive a result.

On write access clients can define to either reject, skip or log intermediate null segments where skip means, that the setting is just silently aborted. Clients can also control how to deal with collection and map intermediates. By default we now propagate the attempt to set a value to all collection or map items respectively. This could be potentially extended to provide a filter receiving both the property and property value to selectively propagate those calls in the future.

The separation of these APIs (`PersistentPropertyAccessor` and `PersistentPropertyPathAccessor`) is introduced as the latter can be implemented on top of the former and the former potentially being dynamically generated. The logic of the latter can then be clearly separated from the actual individual property lookup. `ConvertingPropertyAccessor` is now a `PersistentPropertyPathAccessor`, too, and overrides `SimplePersistentPropertyPathAccessor.getTypedProperty(…)` to plug in conversion logic. This allows custom collection types being used if the `ConversionService` backing the accessor can convert from and to `Collection` or `Map` respectively.

Deprecated all `PersistentPropertyPath`-related methods on `PersistentPropertyAccessor` for removal in 2.3.

`MappingAuditableBeanWrapperFactory` now uses these new settings to skip both null values as well as collection and map values when setting auditing metadata.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.2.0.BUILD-SNAPSHOT</version>
+	<version>2.2.0.DATACMNS-1438-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
--- a/src/main/java/org/springframework/data/auditing/MappingAuditableBeanWrapperFactory.java
+++ b/src/main/java/org/springframework/data/auditing/MappingAuditableBeanWrapperFactory.java
@@ -224,7 +224,7 @@ public class MappingAuditableBeanWrapperFactory extends DefaultAuditableBeanWrap
 			return setDateProperty(metadata.lastModifiedDatePaths, value);
 		}
 
-		/* 
+		/*
 		 * (non-Javadoc)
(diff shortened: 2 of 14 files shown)


pivotal-issuemaster:
@odrotbohm Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=330)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=330) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@odrotbohm Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=330)!

PR #332, opened 2019-01-14
DATACMNS-1467 - Entity Callback API

mp911de:
This draft sketches out a possible Entity Callback API heavily inspired by Spring Framework's Application Event listeners. Entity callbacks are callbacks for entities that allow for entity modification at certain check points such as before saving an entity or after loading it.

Entity callbacks consist of a marker-interface for all entity callback types and the EntityCallbacks API to dispatch callbacks.

Entity callbacks are picked up from the ApplicationContext and invoked sequentially according to their ordering.

Usage example:

```java
AnnotationConfigApplicationContext ctx = new AnnotationConfigApplicationContext(MyConfig.class);

SimpleEntityCallbacks callbacks = new SimpleEntityCallbacks(ctx);

PersonDocument personDocument = new PersonDocument(null, "Walter", null);
PersonDocument afterCallback = callbacks.callback(personDocument, BeforeSaveCallback.class,
		BeforeSaveCallback::onBeforeSave);

Mono<PersonDocument> afterCallback = callbacks.callbackLater(personDocument, ReactiveBeforeSaveCallback.class,
		ReactiveBeforeSaveCallback::onBeforeSave);

@configuration
static class MyConfig {

	@bean
	BeforeSaveCallback<Person> personCallback() {
		return object -> {
			object.setSsn(object.getFirstName().length());
			return object;
		};
	}

	@bean
	MyReactiveBeforeSaveCallback callback() {
		return new MyReactiveBeforeSaveCallback();
	}

}

class MyReactiveBeforeSaveCallback implements ReactiveBeforeSaveCallback<Person> {

	@override
	public Mono<Person> onBeforeSave(Person object) {

		PersonDocument result = new PersonDocument(object.getFirstName().length(), object.getFirstName(),
				object.getLastName());

		return Mono.just(result);
	}
}
```

---

Related ticket: [DATACMNS-1467](https://jira.spring.io/browse/DATACMNS-1467).


src/main/java/org/springframework/data/mapping/callback/BeforeSaveCallback.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Entity callback before saving the entity. This interface allows for modifications to the actual entity that are
+ * required prior to saving the entity.
+ *
+ * @author Mark Paluch
+ * @see EntityCallback
+ * @see SimpleEntityCallbacks
+ */
+@FunctionalInterface
+public interface BeforeSaveCallback<T> extends EntityCallback<T> {
```
mp911de: Probably we should move `BeforeSaveCallback` and `ReactiveBeforeSaveCallback` to test-code only as store events are carrying typically a lot more of auxiliary data (collection name, document).

src/main/asciidoc/entity-callbacks.adoc
```diff
@@ -0,0 +1,163 @@
+[entity-callbacks]
+= Entity Callbacks
+
+The Spring Data infrastructure provides hooks for modifying an entity before and/or after certain methods are invoked.
```
Buzzardo: "and/or" can be "and" here. Indicating that both are possible doesn't mean that both are required, which makes the "or" unnecessary. ("and/or" is hardly ever needed.)

src/main/asciidoc/entity-callbacks.adoc
```diff
@@ -0,0 +1,163 @@
...
+Those so called ``EntityCallback``s provide a convenient way to check and potentially modify an entity in a callback fashioned style. +
```
Buzzardo: Please don't pluralize code objects in this way. Rather. refer to `EntityCallback` instances or `EntityCallback` objects or some similar word. (In a nut shell, use the class or interface name as an adjective for a noun.)

src/main/asciidoc/entity-callbacks.adoc
```diff
@@ -0,0 +1,163 @@
...
+An `EntityCallback` looks pretty much like a specialized `ApplicationListener`, with which you might be already familiar.
```
Buzzardo: You can leave off ", with which you might be already familiar." Whether they are or not, you want them to read the next sentence, so you don't need to mention familiarity. (If you did keep it, it should be ", with with you might already be familiar" - English gets weird about adverbs).

src/main/asciidoc/entity-callbacks.adoc
```diff
@@ -0,0 +1,163 @@
...
+Some Spring Data modules publish store specific events, like a `BeforeSaveEvent` that allow modifying the given entity which in some cases, eg. immutable types, can cause trouble.
```
Buzzardo: I would rewrite this to be "Some Spring Data modules publish store-specific events (such as `BeforeSaveEvent`) that allow modifying the given entity. In some cases, such as when working with immutable types, these events can cause trouble.

src/main/asciidoc/entity-callbacks.adoc
```diff
@@ -0,0 +1,163 @@
...
+Another aspects is that, event publishing relies on `ApplicationEventMulticaster` that can be configured with an asynchronous `TaskExecutor` leading to unpredictable outcome as event processing can be forked onto a `Thread`.
```
Buzzardo: I would rewrite this one to be: "Also, event publishing relies on `ApplicationEventMulticaster`. Beware of configuring that with an asynchronous `TaskExecutor`. Doing so can lead to unpredictable outcomes, as event processing can be forked onto a `Thread`."

src/main/asciidoc/entity-callbacks.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+Entity callbacks provide integration points with both, synchronous and reactive, APIs guaranteeing an in-order execution at well-defined checkpoints within the processing chain, returning a potentially modified entity or an reactive wrapper type.
```
Buzzardo: I would rewrite that as "Entity callbacks provide integration points with both synchronous and reactive APIs to guarantee in-order execution at well-defined checkpoints within the processing chain, returning a potentially modified entity or an reactive wrapper type."

src/main/asciidoc/entity-callbacks.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+Entity callbacks are typically separated by API type. This separation means that a synchronous API only considers synchronous entity callbacks and a reactive implementation considers only reactive entity callbacks.
```
Buzzardo: Change "a synchronous API only considers" to "a synchronous API considers only".

src/main/asciidoc/entity-callbacks.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+[NOTE]
+====
+The Entity Callback API has been introduced with Spring Data Commons 2.2. It is the recommended way of applying entity modifications.
+Existing store specific `ApplicationEvents` are still published *before* the invoking potentially registered ``EntityCallback``s.
```
Buzzardo: "store-specific" should be hyphenated.

src/main/asciidoc/entity-callbacks.adoc
```diff
@@ -0,0 +1,163 @@
...
+====
+
+[entity-callbacks.implement]
+== Implementing Entity Callbacks
+
+An `EntityCallback` is directly associated with its domain type through its generic type argument.
+Each Spring Data module typically ships a set of predefined `EntityCallback` interfaces covering the entity lifecycle.
```
Buzzardo: Insert "with" after "ships".

christophstrobl:
Thanks @Buzzardo!

PR #338, opened 2019-02-19
DATACMNS-1483 - Migrate Kotlin tests to Mockk

sdeleuze:
In order to be consistent with Spring Framework tests and recommendation to use [Mockk](https://mockk.io/) for Kotlin tests, and in order to be consistent with the upcoming Coroutines pull requests for Spring Data MongoDB and R2DBC which are using Mockk as well, I have migrated Spring Data Commons tests and will do as well with other Spring Data projects.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -264,24 +264,10 @@
 		</dependency>
 
 		<dependency>
-			<groupId>com.nhaarman</groupId>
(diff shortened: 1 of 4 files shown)


mp911de:
Thanks a lot. That's merged, polished, and backported now.

PR #341, opened 2019-03-01
DATACMNS-1486 - add missing Nullable annotation

hho:
Fix [DATACMNS-1486](https://jira.spring.io/browse/DATACMNS-1486) by adding the missing `@Nullable`.


mp911de:
Thanks a lot, that's merged and polished now.

PR #342, opened 2019-03-02
DATACMNS-1487 - Dont cache candidate bean definitions

paulcwarren:
By not caching the discovered bean definitions (as the cached discovery is based on base packages of the repository) this fixed [DATACMNS-1487](https://jira.spring.io/browse/DATACMNS-1487) for me.

I consider this a partial PR as this change makes this constructor potentially removable.  However, I couldn't get the tests compiling on my clone due to missing querydsl classes and therefore I was unable to refactor the constructor out of the code and tests.  

You guys may have a better, more complete fix anyways.


odrotbohm:
Rejected for the reasons described in the ticket.

PR #390, opened 2019-04-08
DATACMNS-1509 - Ignore override properties with non-assignable types when creating PersistentEntity.

christophstrobl:
delivers input to: spring-projects/spring-data-mongodb#737


mp911de:

PR #391, opened 2019-04-09
DATACMNS-1512 - Allow customization of ConversionService in RepositoryFactorySupport

wreulicke:
Fix [DATACMNS-1512](https://jira.spring.io/projects/DATACMNS/issues/DATACMNS-1512)

I have a library using spring-data-commons, [spring-data-mirage](https://github.com/dai0304/spring-data-mirage).

In the library, it is able to use custom collection as return type, such as a below custom collection.

We have the interface, Chunk and the implementation, ChunkImpl.
But, We cannot return ChunkImpl as implementation of Chunk in 2.1.6.RELEASE of spring-data-commons.
This causes by element type conversion introduced by https://jira.spring.io/browse/DATACMNS-1482.

```
public interface Chunk<T> extends Collection<T> {
  // some methods...
}

public class ChunkImpl<T> implements Chunk<T> {
  // some implmentations...
}
```

I cannot find the workaround.
So, I would like to make to register custom conversion.

# checklist

### DIFF
--- a/src/main/java/org/springframework/data/repository/core/support/QueryExecutionResultHandler.java
+++ b/src/main/java/org/springframework/data/repository/core/support/QueryExecutionResultHandler.java
@@ -39,6 +39,7 @@ import org.springframework.lang.Nullable;
  * @author Oliver Gierke
  * @author Mark Paluch
  * @author Jens Schauder
+ * @author wreulicke
  */
 class QueryExecutionResultHandler {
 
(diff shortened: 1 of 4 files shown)


mp911de:
It would make sense to have an externally configurable `ConversionService` by introducing a setter instead of delegating converter registration calls.

Beyond that, we could pre-initialize `ConfigurationService` with a shared and pre-configured instance. `QueryExecutionResultHandler` uses `ConfigurationService` without modifying it and we would not create a `ConfigurationService` per repository instance. Probably a separate ticket though.

Wdyt @odrotbohm?

wreulicke:
@mp911de Thanks for your review. 😄 

I implemented to use pre-configured instance in 9c90f83. 
What do you think?

wreulicke:
rebased

odrotbohm:
Please refer to the comments and questions made on the ticket back in May. I don't think we should or can move ahead with this one without the concerns addressed.

PR #395, opened 2019-05-14
DATACMNS-1534 - Cache BeanFactory lookup for EvaluationContextExtension.

christophstrobl:
We now cache the `BeanFactory` lookup for `EvaluationContextExtension` within the `ExtensionAwareEvalutationContextProvider` to avoid (expensive) repeated context scans when creating the actual `EvaluationContext`.


odrotbohm:
Merged and backported to Lovelace.

PR #396, opened 2019-05-21
DATACMNS-1538 - Add support for custom repository proxy post processors in CDI

arielcarrera:
Add support for custom repository proxy post processors in CDI


schauder:
Thanks, that's squashed polished and merged.

PR #401, opened 2019-08-15
DATACMNS-1580 - Updated documentation

hijxf:
Initially, I went to the API to try and see if the method returned a potentially modified `entity`, but I felt the documentation wasn't clear on that. So we went to test it and _obviously_ was the case 😅.

After studying this doc for more than a couple minutes, we can assume it returns a potential modified `entity` based on its return type, but still felt like we needed to update the doc 🤓.

Just my opinion, but hopefully this helps. I also don't feel like I deserve a spot in the authors page, so I didn't add my name even though it's part of the checklist...

Thanks!
Jef

- [ ] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATACMNS).

### DIFF
--- a/src/main/java/org/springframework/data/repository/CrudRepository.java
+++ b/src/main/java/org/springframework/data/repository/CrudRepository.java
@@ -31,7 +31,7 @@ public interface CrudRepository<T, ID> extends Repository<T, ID> {
 	 * entity instance completely.
 	 *
 	 * @param entity must not be {@literal null}.
-	 * @return the saved entity will never be {@literal null}.
+	 * @return the saved entity; will never be {@literal null}.
 	 */
 	<S extends T> S save(S entity);
 
@@ -39,7 +39,7 @@ public interface CrudRepository<T, ID> extends Repository<T, ID> {
 	 * Saves all given entities.
 	 *
 	 * @param entities must not be {@literal null}.
-	 * @return the saved entities will never be {@literal null}.
+	 * @return the saved entities; will never be {@literal null}.
 	 * @throws IllegalArgumentException in case the given entity is {@literal null}.
 	 */
 	<S extends T> Iterable<S> saveAll(Iterable<S> entities);


pivotal-issuemaster:
@hijxf Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=401)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=401) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@hijxf Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=401)!

odrotbohm:
I am puzzled because the Javadoc literally said:

> the save operation might have changed the entity instance completely

We have to be careful about changes here as we have to make sure that de documented behavior matches what the different store implementations actually do. The extended generics are only in place so that you can get a potential sub-type you hand into the method as the subtype itself. I.e. the generics declaration does not have anything to do with the fact, that the method might just return the same instance but it could also return a completely new one, e.g. when dealing with immutable aggregate types.

That said, I'm not convinced the new wording is better than the old, as – in fact – it suggests modification where potentially none happens.

hijxf:
That's a great point. I think the verbage in the description is fair and
clear in the original code. Would you agree adding a semi colon after 'the
saved entity' would be clearer? It's a small enough change as just looking
at the return type can be confusing. I also want to be clear I'm not trying
to split hair here, but just suggest.

Thanks!

hijxf:
Updated the PR with my suggested change. If you don't feel that it's necessary, no problem. Thanks for your time.

mp911de:
That's merged and backported now. Thank you.

hijxf:
Thanks!

PR #402, opened 2019-08-26
DATACMNS-1570 - Made the behaviour consistent between Subject part and Order By part.

schauder:
DATACMNS-1570 - Made the behaviour consistent between Subject part and Order By part.

The subject part wasn't decapitalizing properties starting with multiple characters properly.


schauder:
Merged.

PR #408, opened 2019-10-10
fix projection with EnumSet field

wilds:
When you try to apply a projecton to an entity with some EnumSet field, it fails with "Cannot create EnumSet for unknown element type"

Problem related to: https://github.com/spring-projects/spring-boot/issues/15539


mp911de:
The 1.x line is out of maintenance and we do not intend to ship another 1.x release.

PR #410, opened 2019-10-15
DATACMNS-1591 - @Primary is now considered on repository interfaces

odrotbohm:
We now elevate the primary annotation from the scanned repository bean definition to the one created for the repository factory. This previously got lost as the former is hidden behind the RepositoryConfiguration interface that previously didn't expose the primary nature of the underlying bean definition.


mp911de:

PR #411, opened 2019-10-24
DATACMNS-1596 - Modules not exposing identifying annotations do not claim repository interfaces anymore.

odrotbohm:
Previously, a module not exposing any entity identifying annotations would have claimed a repository definition and potentially overrode the bean definition of another store that was the proper claim in the first place. We have now changed this to abstain from a claim of the interface.

We' also tweaked the log output in the following cases:

1. If the module neither returns an identifying type nor entity identifying annotations, we now log a warning that a module does not support a multi-module setup and answer all assignment requests with false.

2. The info level warning indicating an interface has been dropped now reports which annotations or interface base types to use.

Related tickets: spring-projects/spring-boot#18721

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.3.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


mp911de:

PR #413, opened 2019-10-30
DATACMNS-1597 - Skip non-simple type converter registrations.

Walliee:
If a converter is annotated with `@ReadingConverter` or `@WritingConverter`, then decision on whether to register it can be made based on whether the source/target type is simple or not.

https://jira.spring.io/browse/DATACMNS-1597


mp911de:
Superseded by #421 as per comment on the linked ticket.

PR #414, opened 2019-11-08
DATACMNS-1609 - Auditing metadata should skip paths containing map or collection like segments.

christophstrobl:
Should be back ported to _2.2.x_.


christophstrobl:
Applied [DATACMNS-1438](https://jira.spring.io/browse/DATACMNS-1438) & [DATACMNS-1461](https://jira.spring.io/browse/DATACMNS-1461) (both already present in 2.1.x) to 2.2.x and master.

PR #416, opened 2019-11-18
DATACMNS-1618 - Set log level for potentially unsupported custom conversion java.time target type to debug.

christophstrobl:
This is a temporary solution that seeks a better one in 2.3 and therefore should not be forward ported!


mp911de:
That's merged now.

PR #420, opened 2019-11-20
DATACMNS-1625 - Lazily initialize EvaluationContextExtension functions and properties.

christophstrobl:
We now make sure to lazily evaluate `Functions` and properties of a given `EvaluationContextExtension` based on the `rootObject` that might not be present when creating the context in first place, but delay the resolution to a time that actually may call the desired function.

----

Should be forward ported to 2.2.x and 2.3

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.1.16.BUILD-SNAPSHOT</version>
+	<version>2.1.16.DATACMNS-1625-SNAPSHOT</version>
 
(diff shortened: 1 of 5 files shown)


odrotbohm:
I'd love to see a bit more elaboration in the commit message on the changes to the internals. Why the move to `AtomicBoolean`? Why the `ConcurrentSkipListMap`? What's the `Properties` implementation doing on a high level?

To the unknowing eye it looks like there's some memoization implemented while `Lazy` already memoizes the calculated value.

mp911de:
Should be no longer necessary once #454 is merged.

mp911de:
Closing without merge in favor of #454.

PR #421, opened 2019-11-22
DATACMNS-1615 - Allow fine grained store specific converter registration.

christophstrobl:
`CustomConversions` now registers all given user defined converters and selects only those converters from the `StoreConversions` (potentially containing superfluous default converters from  spring-data-commons) that convert to or from a store supported simple type.
By doing so we make sure to only register supported converters which removes warning messages from the log and reduces the number of converters within the `ConversionService`.

### DIFF
--- a/src/main/java/org/springframework/data/convert/CustomConversions.java
+++ b/src/main/java/org/springframework/data/convert/CustomConversions.java
@@ -25,6 +25,7 @@ import lombok.extern.slf4j.Slf4j;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.springframework.core.GenericTypeResolver;
 import org.springframework.core.convert.converter.Converter;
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/convert/CustomConversions.java
```diff
@@ -57,6 +58,10 @@
 	private static final String READ_CONVERTER_NOT_SIMPLE = "Registering converter from %s to %s as reading converter although it doesn't convert from a store-supported type! You might wanna check you annotation setup at the converter implementation.";
 	private static final String WRITE_CONVERTER_NOT_SIMPLE = "Registering converter from %s to %s as writing converter although it doesn't convert to a store-supported type! You might wanna check you annotation setup at the converter implementation.";
 	private static final String NOT_A_CONVERTER = "Converter %s is neither a Spring Converter, GenericConverter or ConverterFactory!";
+	private static final String CONVERTER_FILTER = "converter from %s to %s as %s converter.";
+	private static final String ADD_CONVERTER = "Adding %s" + CONVERTER_FILTER;
+	private static final String SKIP_CONVERTER = "Skipping " + CONVERTER_FILTER
+			+ "%s is not a store supported simple type!";
```
martin-g: This probably needs a leading empty space character.

christophstrobl: 👍 thanks @martin-g !

mp911de:

PR #424, opened 2019-12-13
KotlinExt. Fix generic type of ID to be not null

haufau:
`public interface CrudRepository<T, ID> extends Repository<T, ID> {`

...

`/**`
	 `* Retrieves an entity by its id.`
	` *`
	` * @param id must not be {@literal null}.`
	 `* @return the entity with the given id or {@literal Optional#empty()} if none found.`
	 `* @throws IllegalArgumentException if {@literal id} is {@literal null}.`
	 `*/`
	`Optional<T> findById(ID id);`

...
`}`

`findById` method requires not null id parameter. So, in Kotlin extension we can forbid nullability in generic type as `null` is not a subclass of `Any`, but of `Any?` which wildcard generic is.

### DIFF
--- a/src/main/kotlin/org/springframework/data/repository/CrudRepositoryExtensions.kt
+++ b/src/main/kotlin/org/springframework/data/repository/CrudRepositoryExtensions.kt
@@ -23,4 +23,4 @@ package org.springframework.data.repository
  * @author Sebastien Deleuze
  * @since 2.1.4
  */
-fun <T, ID> CrudRepository<T, ID>.findByIdOrNull(id: ID): T? = findById(id).orElse(null)
+fun <T, ID: Any> CrudRepository<T, ID>.findByIdOrNull(id: ID): T? = findById(id).orElse(null)


pivotal-issuemaster:
@haufau Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=424)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=424) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@haufau Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-commons&pullRequestId=424)!

mp911de:
Care to elaborate? The current code causes a compile error when calling `repository.findByIdOrNull(null)` so the proposed change does not seem necessary.

mp911de:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

haufau:
Are you sure about compile error? I pretty sure that i can pass nullable type to it. At least at 2.2.4.

Type in current type-parameter declaration not forbid null value.

mp911de:
I used Kotlin 1.3 to verify the functionality.

haufau:
Okay, it was not so simple, but I found it. If you extend `CrudRepository<T, ID>`,  it's compile error. But when you extend `JpaRepository` which extends `CrudRepository` through `PagingAndSortingRepository` it becomes `ID!` type, means it ignores null-safety

mp911de:
Isn't that an issue on the Kotlin side? `PagingAndSortingRepository` isn't special in any regard to generics and I'd expect Kotlin to handle the issue properly. Is there already a ticket in the Kotlin issue tracker we could link to?

haufau:
Seems like if we create any `.java` interface extending `CrudRepository`, and extend that one with `.kt` interface information about type will be erased. Not affecting if we do `.kt` interface chain exactly from `CrudRepository`. I already asked in community is it bug, or feature.

But also is it not correct to fix this explicit with `ID: Any` on extension, to make not possible further runtime error on `findById` not null assertion?

PR #425, opened 2019-12-28
DATACMNS-1641 - Allow auditing annotations to configure overwrite behavior

daniel-shuy:
There are scenarios where we want the `@CreatedBy` and `@CreatedDate` auditing annotations to be ignored if the existing value is not null, eg. when synchronizing events/entities from one service/database to another (which can be quite a common pattern when using the databases as immutable event stores in a microservice architecture, where events are published to a messaging bus, then consumed asynchronously and cached in the dependent services' database). Currently the only workaround is to create custom entity listeners.

This PR adds an `overwriteBehavior` parameter to `@CreatedBy` and `@CreatedDate`. The parameter defaults to `OverwriteBehavior#OVERWRITE`, which is the same as the current behavior (for backwards compatibility), but can be set to `OverwriteBehavior#SKIP` to not overwrite the existing value if it is not `null`, eg.
```java
public class Auditable {
    @CreatedBy(overwriteBehavior = OverwriteBehavior.SKIP)
    private String createdBy;

    @CreatedDate(overwriteBehavior = OverwriteBehavior.SKIP)
    private Instant createdDate;
}
```

To maintain consistency, I also enhanced `DefaultAuditableBeanWrapperFactory.AuditableInterfaceBeanWrapper` to allow the overwrite behavior for entity classes implementing the [Auditable](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Auditable.html) interface to be configured by annotating the getter methods ([getCreatedBy()](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Auditable.html#getCreatedBy--) or [getCreatedDate()](https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/domain/Auditable.html#getCreatedDate--)) with the corresponding annotation (`@CreatedBy` or `@CreatedDate`), eg.
```java
@MappedSuperclass
public abstract class AuditableImpl implements AbstractAuditable<String, Long> {
    @CreatedBy(overwriteBehavior = OverwriteBehavior.SKIP)
    @Override
    public Optional<String> getCreatedBy() {
        return super.getCreatedBy();
    }

    @CreatedDate(overwriteBehavior = OverwriteBehavior.SKIP)
    @Override
    public Optional<LocalDateTime> getCreatedDate() {
        return super.getCreatedDate();
    }
}
```

### DIFF
--- a/src/main/java/org/springframework/data/annotation/CreatedBy.java
+++ b/src/main/java/org/springframework/data/annotation/CreatedBy.java
@@ -21,14 +21,23 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import org.springframework.data.auditing.OverwriteBehavior;
+
 /**
  * Declares a field as the one representing the principal that created the entity containing the field.
  *
  * @author Ranie Jade Ramiso
  * @author Oliver Gierke
+ * @author Daniel Shuy
  * @since 1.5
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(value = { FIELD, METHOD, ANNOTATION_TYPE })
 public @interface CreatedBy {
+
+	/**
+	 * The behavior if existing field value is not {@code null}. Defaults to {@link OverwriteBehavior#OVERWRITE}.
+	 */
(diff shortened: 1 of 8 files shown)


mp911de:
Closing as per comment in the [associated ticket](https://jira.spring.io/browse/DATACMNS-1641).

PR #432, opened 2020-02-28
DATACMNS-1676 - Create shared conversion service for AuditableBeanWrapper

pavelhoral:
While profiling our application I have discovered that a significant amount of time is being spent in conversion service instantiation (image is from 20s HTTP request):

image

Creating conversion service for each persisted object was not necessary so I have moved it to a static initializer. This is a similar change to https://jira.spring.io/browse/DATACMNS-1357 .

I have also tried to get rid of `DefaultFormattingConversionService` (replace with `GenericConversionService`) as it seemed a bit unecessary, but unfortunately [two of it's package private converters](https://github.com/spring-projects/spring-framework/blob/master/spring-context/src/main/java/org/springframework/format/datetime/standard/DateTimeConverters.java#L75) are used when converting from/to long.

---

> You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

Not sure if I should add myself when this is such a minor change (kind of "obvious fix" category).

### DIFF
--- a/src/main/java/org/springframework/data/auditing/DefaultAuditableBeanWrapperFactory.java
+++ b/src/main/java/org/springframework/data/auditing/DefaultAuditableBeanWrapperFactory.java
@@ -21,7 +21,6 @@ import lombok.RequiredArgsConstructor;
 import java.lang.reflect.Field;
 import java.time.Instant;
 import java.time.temporal.TemporalAccessor;
-import java.util.Calendar;
 import java.util.Date;
 import java.util.Optional;
 import java.util.stream.Stream;
@@ -151,7 +150,7 @@ class DefaultAuditableBeanWrapperFactory implements AuditableBeanWrapperFactory 
 			return value;
 		}
 
-		/* 
+		/*
 		 * (non-Javadoc)
 		 * @see org.springframework.data.auditing.AuditableBeanWrapper#getBean()
 		 */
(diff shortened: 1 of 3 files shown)


pavelhoral:
I have also replace mentions of `Calendar` in JavaDoc with `TemporalAccessor` as the old Date API is no longer used by that class.

odrotbohm:
This definitely needs a fix and I like where this is heading. That said, I'd prefer to avoid the static instance of the `ConversionService`. Would you mind to try to rather create the shared service instance in the Factory creating the BeanWrappers and hand them into the constructors creating them? That would allow us to keep them in a member field.

pavelhoral:
> Would you mind to try to rather create the shared service instance in the Factory creating the BeanWrappers and hand them into the constructors creating them?

Will do... might take me couple days as I am on my vacation at the moment.

odrotbohm:
No worries, no need to rush. I might find time next week and will report back. Please check back here before you invest further time. Enjoy the time off!

pavelhoral:
* Created second commit so that the changes are more apparent. If you like the changes, I can squash it.
* Added new static conversion service factory method `DefaultAuditableBeanWrapperFactory#getDateConversionService` such that other components and unit tests can ask for it.
* I chose to have conversion service as the first constructor parameter as it felt a bit more natural. If you want it to be the last parameter, let me know.

mp911de:

PR #433, opened 2020-03-10
DATACMNS-1678 - Implemented ComparatorUtils.

konstiak:


mp911de:
After reiterating here, we couldn't convince ourselves to add your contribution. Probably there should be a separate place outside of Spring Data where this extension could live in.

PR #434, opened 2020-03-16
DATACMNS-1683 - Rebuild quotation index from rewritten SpEL query

mp911de:
We now use the correct quotation map that is based on the rewritten SpEL query to detect whether an expression was quoted. Previously, we used the quotation map from the original query. After augmenting the query with synthetic parameters, the quotation offset no longer matched the query that was under inspection and calls to `SpelExtractor.isQuoted(…)` could report an improper result.

---

Related ticket: [DATACMNS-1683](https://jira.spring.io/browse/DATACMNS-1683).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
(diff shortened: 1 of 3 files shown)


schauder:
That's merged and back ported. Thanks

PR #440, opened 2020-04-27
DATACMNS-1710 - Avoid potentially expensive constructor parameter creation where possible

christophstrobl:
Reorder constructor filtering with the goal to delay/avoid potentially expensive `PreferredConstructor` creation with parameter lookup as much as possible.

The lookup is (in a default setup) only done once per domain type, but depending on the number of entities to inspect the change can help save some cycles for larger domains.

**Before**
```bash
Benchmark                                  Mode  Cnt         Score         Error  Units
EntityMetadataBenchmark.alwaysNew         thrpt   10    224318,163 ±   42542,453  ops/s
```

**After**
```bash
Benchmark                                  Mode  Cnt         Score         Error  Units
EntityMetadataBenchmark.alwaysNew         thrpt   10    420053,505 ±    9288,093  ops/s
```

```java
public class DomainType {

	private String id;
	private @id String theId;

	private String firstname, lastname;
	private Integer age;

	public DomainType(String id, String theId, String firstname, String lastname, Integer age) {
		this.id = id;
		this.theId = theId;
		this.firstname = firstname;
		this.lastname = lastname;
		this.age = age;
	}

	@PersistenceConstructor
	public DomainType(String firstname, String lastname) {

		this.firstname = firstname;
		this.lastname = lastname;
	}
}
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.3.0.BUILD-SNAPSHOT</version>
+	<version>2.3.0.DATACMNS-1710-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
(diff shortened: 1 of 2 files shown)


mp911de:

PR #443, opened 2020-05-11
DATACMNS-1721 - Lazily evaluate potential property matches in PropertyReferenceException.

christophstrobl:
In some cases the caller of a `PropertyPath` is not interested in the potentially matching properties provided via `PropertyReferenceException`. 
Those cases would benefit from a lazy evaluation.


mp911de:
That's merged now.

PR #447, opened 2020-05-29
DATACMNS-1736 - Use BeanFactory lookup in configuration classes instead of intercepted local bean methods

mp911de:
We now use `@Configuration(proxyBeanMethods = false)` and use `BeanFactory` bean lookups to avoid CGlib proxy creation.

Using Spring's `BeanProvider.getIfUnique(…)` instead of catching `NoSuchBeanDefinitionException`.

---

Related ticket: DATACMNS-1736

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.4.0-SNAPSHOT</version>
+	<version>2.4.0-DATACMNS-1736-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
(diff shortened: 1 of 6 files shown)


wwadge:
Small aside: this PR is also needed for GraalVM to function without a lot of hacks

mp911de:
That's exactly why we're exploring this option. Likely, we will go ahead and introduce method-injection instead of lookup methods for the next version and discard this PR.

gregturn:
@odrotbohm @mp911de 

I pushed another commit that pivots away from protected methods and instead, when needed, has a protected `Lazy` that can be overridden in subclasses.

wdyt?

gregturn:
I'd love to polish this up and merge to master branch. Then I can consume these changes in Spring Data REST.

PR #450, opened 2020-06-24
Don't consider CompletableFuture as reactive type for repository determination.

meistermeier:
I completely dislike the duplication of the `isDeferred()` check but have not found a suitable alternative to reflect the same behaviour in both classes.


mp911de:

PR #453, opened 2020-06-29
[DATACMNS-1758] Fix ConcurrencyModificationException in DefaultEntityCallbacks

mhyeon-lee:
resolves: [DATACMNS-1758](https://jira.spring.io/browse/DATACMNS-1758)

### DIFF
--- a/src/main/java/org/springframework/data/mapping/callback/EntityCallbackDiscoverer.java
+++ b/src/main/java/org/springframework/data/mapping/callback/EntityCallbackDiscoverer.java
@@ -369,7 +369,7 @@ class EntityCallbackDiscoverer {
 
 		private final Set<EntityCallback<?>> entityCallbacks = new LinkedHashSet<>();
 
-		private final List<EntityCallback<?>> cachedEntityCallbacks = new ArrayList<>();
+		private volatile List<EntityCallback<?>> cachedEntityCallbacks = new ArrayList<>();
 
 		private final Set<String> entityCallbackBeans = new LinkedHashSet<>();
 
@@ -387,14 +387,10 @@ class EntityCallbackDiscoverer {
 
 					List<EntityCallback<?>> entityCallbacks = new ArrayList<>(this.entityCallbacks);
 					AnnotationAwareOrderComparator.sort(entityCallbacks);
-
-					synchronized (cachedEntityCallbacks) {
-						cachedEntityCallbacks.clear();
-						cachedEntityCallbacks.addAll(entityCallbacks);
-					}
+					this.cachedEntityCallbacks = entityCallbacks;
 				}
 
-				return new ArrayList<>(cachedEntityCallbacks);
+				return this.cachedEntityCallbacks;
 			}
 
 			List<EntityCallback<?>> allCallbacks = new ArrayList<>(
--- a/src/main/java/org/springframework/data/mapping/callback/EntityCallbackDiscoverer.java
+++ b/src/main/java/org/springframework/data/mapping/callback/EntityCallbackDiscoverer.java
@@ -394,7 +394,7 @@ class EntityCallbackDiscoverer {
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/mapping/callback/EntityCallbackDiscoverer.java
```diff
@@ -394,7 +394,7 @@ private BeanFactory getRequiredBeanFactory() {
 					}
 				}
 
-				return cachedEntityCallbacks;
+				return new ArrayList<>(cachedEntityCallbacks);
```
mhyeon-lee, line 397: I fixed it again.

I think it's safer to handle it where it returns the result.

src/main/java/org/springframework/data/mapping/callback/EntityCallbackDiscoverer.java
```diff
@@ -387,14 +387,10 @@ private BeanFactory getRequiredBeanFactory() {
...
 					AnnotationAwareOrderComparator.sort(entityCallbacks);
-
-					synchronized (cachedEntityCallbacks) {
-						cachedEntityCallbacks.clear();
-						cachedEntityCallbacks.addAll(entityCallbacks);
-					}
+					this.cachedEntityCallbacks = entityCallbacks;
 				}
 
-				return new ArrayList<>(cachedEntityCallbacks);
```
mhyeon-lee: `new ArrayList` also has a problem.
If `cachedEntityCallbacks.clear()` is executed on another thread before this line is executed, an emptyList is returned.

I think it is easier to change the new cache list. (Give up the final)

mp911de: In that case, it's probably better to roll back the entire optimization to a state where it has been before. That way, we omit all concurrency issues in the first place.

mhyeon-lee: @mp911de 
This is not an optimization, but a fix to solve the concurrency issue.
getEntityCallbacks  method returns `cachedEntityCallbacks` and DefaultEntityCallbacks use it with iterator

https://github.com/spring-projects/spring-data-commons/pull/453/commits/f9839c85e587d3d393ebfe8a6a0152553b49e13b#diff-f9f021397523461fe28f39d98bb39f39L83


However, cachedEntityCallbacks can be modified here. (clear, addAll)
This throws ConcurrentModificationException.

```
Exception:java.util.ConcurrentModificationException: null
  at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1042)
  at java.base/java.util.ArrayList$Itr.next(ArrayList.java:996)
  at org.springframework.data.mapping.callback.DefaultEntityCallbacks.callback(DefaultEntityCallbacks.java:83)
  at org.springframework.data.jdbc.core.JdbcAggregateTemplate.triggerBeforeSave(JdbcAggregateTemplate.java:414)
  at org.springframework.data.jdbc.core.JdbcAggregateTemplate.store(JdbcAggregateTemplate.java:335)
  at org.springframework.data.jdbc.core.JdbcAggregateTemplate.update(JdbcAggregateTemplate.java:183)
```


Addresses concurrency issues by replacing the entire cachedEntityCallbacks without changing them when there are changes.

mp911de: I'm referring to 461647403d64d66cbe70d58ef56eaab4e7d91da3.

mhyeon-lee: @mp911de 
I understood your words. Thank you.

Should I close this PR?

mp911de: Yes, it makes sense that we fix our own issue.

PR #456, opened 2020-07-13
DATACMNS-1767 - Disable ClassGeneratingEntityInstantiator when runnin native-image.

christophstrobl:
Always fall back to reflection when the system property for `org.graalvm.nativeimage.imagecode` is set.
This is required because on the fly code generation is not supported in this case.


mp911de:

PR #459, opened 2020-07-16
DATACMNS-1762 - Support Optional wrapping for projection getters

mp911de:
We now support nullable wrappers for projection interfaces. Getters are inspected whether their return type is a supported nullable wrapper. If so, then the value can be wrapped into that type. Null values default in that case to their corresponding empty wrapper representation.

```java
class Customer {

	Long id;
	String firstname, lastname;
	Address address;
}

interface CustomerProjection {

	Optional<String> getFirstname();

	Optional<AddressExcerpt> getAddress();
}


Customer customer = new Customer();
customer.firstname = "Walter";

CustomerProjection projection = …;

assertThat(projection.getFirstname()).hasValue("Walter");


customer.firstname = null;
assertThat(projection.getFirstname()).isEmpty();
```

Also, extracted `NullableWrapperConverters` from `QueryExecutionConverters` and removed `ReactiveWrappers` from `QueryExecutionConverters `.

---

We should reconsider the package for `NullableWrapperConverters` as further extensions (`Iterable -> Streamable`) might be future enhancements.

Related ticket: DATACMNS-1762
Previous pull request: #457.

### DIFF
--- a/src/main/java/org/springframework/data/util/NullableWrapperConverters.java
+++ b/src/main/java/org/springframework/data/util/NullableWrapperConverters.java
@@ -494,7 +494,7 @@ public abstract class NullableWrapperConverters {
 		}
 	}
 
-	public static final class WrapperType {
+	private static final class WrapperType {
 
 		private WrapperType(Class<?> type, Cardinality cardinality) {
 			this.type = type;
--- a/src/main/java/org/springframework/data/util/NullableWrapperConverters.java
+++ b/src/main/java/org/springframework/data/util/NullableWrapperConverters.java
@@ -1,0 +1,579 @@
+/*
+ * Copyright 2014-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.util;
+
+import scala.Function0;
+import scala.Option;
+import scala.runtime.AbstractFunction0;
+
+import java.util.Collections;
+import java.util.HashSet;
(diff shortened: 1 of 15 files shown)


src/test/java/org/springframework/data/projection/ProxyProjectionFactoryUnitTests.java
```diff
@@ -218,6 +221,61 @@ void doesNotCreateWrappingProxyIfTargetImplementsProjectionInterface() {
...
+
+		CustomerWithOptional excerpt = factory.createProjection(CustomerWithOptional.class, customer);
+
+		assertThat(excerpt.getPicture()).hasValueSatisfying(bytes -> {
+			assertThat(bytes).isEqualTo(new byte[] { 1, 2, 3 });
+		});
+	}
+
+	@Test // DATACMNS-1762
+	void supportsOptionalBackedByOptional() {
```
wimdeblauwe, line 249: Interesting test! Did not think about this use case 👍

src/main/java/org/springframework/data/util/NullableWrapperConverters.java
```diff
@@ -0,0 +1,579 @@
...
+
+			if (source instanceof io.vavr.control.Option) {
+				return ((io.vavr.control.Option<Object>) source).getOrElse(() -> null);
+			}
+
+			return source;
+		}
+	}
+
+	public static final class WrapperType {
```
wimdeblauwe, line 497: This is probably an exact copy of `QueryExecutionConverters.WrapperType`. Shouldn't this be refactored to use the same class? Or is that not possible for backwards-compatibility reasons?

wimdeblauwe:
Thanks for this @mp911de ! I have added 2 remarks above from having a look at what you have done. I could not have done it myself, so I'm very grateful for this 🙇

src/main/java/org/springframework/data/util/NullableWrapperConverters.java
wimdeblauwe: Ok, clear! Looking forward to using this in a future release :-)

NaceHudobivnik:
@mp911de This PR also changed how we get ConversionService in the `ProxyProjectionFactory.java`. Now that you create a new instance, it is no longer possible to add custom converters, so I have no way of adding a converter for Timestamp -> OffsetDateTime, which is currently not supported as a projection getter type. 

How would you propose this could be fixed? I could contribute, but would appreciate some support :)

PR #460, opened 2020-07-18
DATACMNS-1754 - Support implementation lookup for nested repositories and fragments

kirmerzlikin:
**Problem:** while repository and fragment interfaces are discovered correctly (provided that considerNestedRepositories=true), their implementations are not found/registered despite proper class naming.

**Cause:** `DefaultImplementationLookupConfiguration` works in such way that for nested interface (i.e. one whose class name includes name of the enclosing class) expected implementation class name is built including that enclosing class name. In the same time actual implementation class names are always "localized" (i.e. stripped from enclosing class name, if any) before matching. This makes matching implementation classes against nested interface impossible.

**Solution:** when building expected implementation class name, use "local" interface class name, so that it can match any implementation class that follows naming convention "SimpleInterfaceName" + "ImplemenetationPostfix". 
Also for consistency expected implementation bean name now uses "local" interface class name (which additionally allows to avoid having "." in the bean name).

**Results**:
- nothing changes for top-level (not nested) repositories and fragments (because "localizing" their class names doesn't do anything);
- nothing changes in terms of discovery for nested repository and fragment _interfaces_ (because `DefaultImplementationLookupConfiguration` handles already discovered interfaces and implementation classes);
- _implementations_ can now be discovered and registered for nested repository and fragment interfaces.

### DIFF
--- a/src/main/java/org/springframework/data/repository/config/DefaultImplementationLookupConfiguration.java
+++ b/src/main/java/org/springframework/data/repository/config/DefaultImplementationLookupConfiguration.java
@@ -33,6 +33,7 @@ import org.springframework.util.ClassUtils;
  *
  * @author Oliver Gierke
  * @author Mark Paluch
+ * @author Kyrylo Merzlikin
  * @since 2.1
  */
 class DefaultImplementationLookupConfiguration implements ImplementationLookupConfiguration {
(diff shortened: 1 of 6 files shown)


src/main/java/org/springframework/data/repository/config/DefaultImplementationLookupConfiguration.java
```diff
@@ -55,8 +56,7 @@
 
 		this.config = config;
 		this.interfaceName = interfaceName;
-		this.beanName = Introspector
-				.decapitalize(ClassUtils.getShortName(interfaceName).concat(config.getImplementationPostfix()));
+		this.beanName = Introspector.decapitalize(getLocalName(interfaceName).concat(config.getImplementationPostfix()));
```
mp911de, line 59: We should keep the implementation bean name as-is. Switching to the local name can clash with other nested beans and it introduces an inconsistency with fragment interfaces and nested repository bean names that are qualified with their enclosing class.

We can sort out that issue during the merge.

mp911de:

PR #462, opened 2020-08-03
DATACMNS-1780 - PersistentEntities allows in flight metadata creation if the associated mapping context can be identified

christophstrobl:
Use a more lenient approach, that allows metadata creation, when looking up persistent entities. This allows eg. a configured `AuditingHandler` to kick in without having to register an initial entity set in first place.


mp911de:

PR #465, opened 2020-09-06
DATACMNS-1142 - Consider @Primary repository in Repositories if context is given.

janzeppenfeld:
In most of the current cases an ApplicationContext is passed to the Repositories constructor which is a ListableBeanFactory but not a ConfigurableListableBeanFactory. We now check whether the given ListableBeanFactory is of type ConfigurableApplicationContext as well because all current contexts implement this interface and therefore can return a ConfigurableListableBeanFactory. This way we can retrieve the bean definition for a repository and can check whether it has been marked as a primary one.

Related tickets: DATACMNS-1448, DATACMNS-1591, DATAREST-923.


mp911de:
Thanks a lot. We're going to apply a slightly simplified variant of the fix while retaining your test contribution.

mp911de:

janzeppenfeld:
Thanks for the feedback and glad to give something back.

PR #469, opened 2020-10-04
DATACMNS-1808 - Update getPredicate() to support multiple Ops

deepesh-verma:
- `QuerydslPredicateArgumentResolver` calls the `QuerydslPredicateBuilder.getPredicate()` method to create predicates from the arguments.
- Enhance it to support the below request parameters, and allow user the pass the `com.querydsl.core.types.Ops` values as below -
```
?name=deepesh,EQ&age=25,GT&language=[hindi,english],IN
```


JonasXPX (approved):

schauder:
We won't implement this, since we don't want to get in the business of defining query languages.

`QuerydslBinderCustomizer` may be used to customise the existing behaviour to your liking.

PR #470, opened 2020-10-14
DATACMNS-1814 - link fix

Buzzardo:
Fixes the problem of the "Using filters" link jumping to the top of
the document. There was another of these ("Configuration") later in
the document, and this commit fixes that one, too.

See: DATACMNS-1814.


schauder:
Thanks, that's merged.

PR #472, opened 2020-10-28
DATACMNS-1820 - Moves PageabeExecutionUtils to o.s.data.support.

schauder:
This avoids module cycles when it is used from a template.
The old version is still there, deprecated and delegates to the new version.

DATACMNS-1820


schauder:
Merged.

PR #473, opened 2020-11-09
DATACMNS-1827 - Pageable and Sort arguments with empty qualifier value shouldn't be prefixed with delimiter.

vpavic:
This resolves [DATACMNS-1827](https://jira.spring.io/browse/DATACMNS-1827).


mp911de:

vpavic:
@mp911de, I just noticed that you polish commit considered `MergedAnnotations` only for `PageableHandlerMethodArgumentResolverSupport` but not `SortHandlerMethodArgumentResolverSupport` - was that intentional?

mp911de:
Meh, I missed `SortHandlerMethodArgumentResolverSupportˋ. Do you want to submit a PR formten remaining one or should I take care of it?

vpavic:
I can handle it later in the day, I assume it's not that urgent as `2.4.1` seems to have went out few hours ago?
Also, having in mind that'll be targeted at `2.4.2`, we need a new JIRA issue as well, right?

mp911de:
We can reuse the same issue and add 2.4.2 as fix version.

PR #474, opened 2020-11-11
DATACMNS-1827 - Consider Qualifier meta-annotated Sort method parameters.

vpavic:
This is a follow-up on #473 that resolves [DATACMNS-1827](https://jira.spring.io/browse/DATACMNS-1827).

### DIFF
--- a/src/main/java/org/springframework/data/web/SortHandlerMethodArgumentResolverSupport.java
+++ b/src/main/java/org/springframework/data/web/SortHandlerMethodArgumentResolverSupport.java
@@ -22,14 +22,15 @@ import java.util.List;
 import java.util.Optional;
 import java.util.function.Consumer;
 
-import javax.annotation.Nullable;
-
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.core.MethodParameter;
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/web/SortHandlerMethodArgumentResolverSupport.java
```diff
@@ -293,6 +295,19 @@ static boolean notOnlyDots(String source) {
 		return StringUtils.hasText(source.replace(".", ""));
 	}
 
+	@Nullable
+	private static String getQualifier(@Nullable MethodParameter parameter) {
```
vpavic, line 299: @mp911de, since this piece of code duplicates the equally named method from `PageableHandlerMethodArgumentResolverSupport`, we could consider moving it to some utility class.

Not sure whether `SpringDataAnnotationUtils` is the right place, due to its (possibly outdated) javadoc?

mp911de: Yes, it makes sense. I'll take care of this during the merge.

mp911de:

PR #475, opened 2020-11-11
Add the required override to the example.

michael-simons:
The snipped does not compile as is. This addition makes it easier to understand what is actually required.


mp911de:

PR #478, opened 2020-11-26
DATACMNS-1835 - Provide Type-safe Kotlin query extension

mp911de:
We now provide a `KPropertyPath` extension leveraging `KProperty` references to express a property path. Using Kotlin property references (such as `Author::name` that translates to `book` or `Book::author / Author.name` translating to `book.author`) is refactoring-safe as the property expressions are part of the Kotlin language. We render `KPropertyPath` using the same semantics as our `PropertyPath` so that store-specific modules can leverage either property paths or accept `KProperty` directly.

Related ticket: DATACMNS-1835.

### DIFF
--- a/src/main/kotlin/org/springframework/data/mapping/KPropertyPathExtensions.kt
+++ b/src/main/kotlin/org/springframework/data/mapping/KPropertyPathExtensions.kt
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


src/main/kotlin/org/springframework/data/mapping/KPropertyPathExtensions.kt
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.mapping
+
+import kotlin.reflect.KProperty
+
+/**
+ * Extension for [KProperty] providing an `toPath` function to render a [KProperty] in dot notation.
```
sdeleuze, line 21: `toDotPath` function?

sdeleuze:
:+1: for this PR.

PR #479, opened 2020-12-07
DATACMNS-1836 - Adds JSR310 converters to the converters used for projections.

schauder:
Similar converters where removed because they clashed with the support for nullable wrappers. This adds them back as dedicated converters that does not clash with conversions for nullable wrappers.

See bd3992d#diff-96f7084777a2a702d85f5c600e542b10123d6c3332ed3d3f8e329ca7061faab0R55

See DATACMNS-1836

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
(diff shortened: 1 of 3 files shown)


schauder:
The PR for this is https://github.com/spring-projects/spring-data-commons/pull/479

PR #126, opened 2015-06-02
1.9.x - PropertyPath is not dealing with properties containing underscore

sabinewinkler:
Having an entity with an underscore property e.g. "first_name" and a related repository method "findByFirst__name" (as documented with double underscore - http://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.query-property-expressions) results in the following exception:

    java.lang.IllegalStateException: Failed to load ApplicationContext
	at ....
    Caused by: org.springframework.data.mapping.PropertyReferenceException: No property first found for type Todo!
	at org.springframework.data.mapping.PropertyPath.<init>(PropertyPath.java:75)
	at org.springframework.data.mapping.PropertyPath.create(PropertyPath.java:327)
	at org.springframework.data.mapping.PropertyPath.create(PropertyPath.java:307)
	at org.springframework.data.mapping.PropertyPath.from(PropertyPath.java:270)
	at org.springframework.data.mapping.PropertyPath.from(PropertyPath.java:241)
	at org.springframework.data.repository.query.parser.Part.<init>(Part.java:76)
	at org.springframework.data.repository.query.parser.PartTree$OrPart.<init>(PartTree.java:235)
	at org.springframework.data.repository.query.parser.PartTree$Predicate.buildTree(PartTree.java:373)
	at org.springframework.data.repository.query.parser.PartTree$Predicate.<init>(PartTree.java:353)
	at org.springframework.data.repository.query.parser.PartTree.<init>(PartTree.java:87)
	at org.springframework.data.jpa.repository.query.PartTreeJpaQuery.<init>(PartTreeJpaQuery.java:61)
	at org.springframework.data.jpa.repository.query.JpaQueryLookupStrategy$CreateQueryLookupStrategy.resolveQuery(JpaQueryLookupStrategy.java:94)
	at org.springframework.data.jpa.repository.query.JpaQueryLookupStrategy$CreateIfNotFoundQueryLookupStrategy.resolveQuery(JpaQueryLookupStrategy.java:205)
	at org.springframework.data.jpa.repository.query.JpaQueryLookupStrategy$AbstractQueryLookupStrategy.resolveQuery(JpaQueryLookupStrategy.java:72)
	at org.springframework.data.repository.core.support.RepositoryFactorySupport$QueryExecutorMethodInterceptor.<init>(RepositoryFactorySupport.java:349)
	at org.springframework.data.repository.core.support.RepositoryFactorySupport.getRepository(RepositoryFactorySupport.java:187)
[...]

PR #480, opened 2020-12-15
DATACMNS-1699 - Add Embedded annotation.

christophstrobl:
Add core annotation for embedded object support and consider a potential prefix when using `FieldNamingStrategies`.

The actual store specific implementation is done in the individual modules that make sure the mapping context provides means to detect embedded types and provide this information to the `PersistentPropertyPathFactory` now using` `PersistentEntities` instead of the `TypeInformation` to construct mapped paths.

### DIFF
--- a/src/main/java/org/springframework/data/mapping/PersistentProperty.java
+++ b/src/main/java/org/springframework/data/mapping/PersistentProperty.java
@@ -16,6 +16,7 @@
 package org.springframework.data.mapping;
 
 import java.lang.annotation.Annotation;
+import java.lang.annotation.ElementType;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.Collection;
@@ -23,6 +24,9 @@ import java.util.Map;
 
 import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.core.annotation.AnnotationUtils;
+import org.springframework.data.annotation.Embedded;
+import org.springframework.data.annotation.Embedded.OnEmpty;
+import org.springframework.data.util.NullableUtils;
 import org.springframework.data.util.TypeInformation;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
@@ -33,6 +37,7 @@ import org.springframework.util.Assert;
  * @author Oliver Gierke
  * @author Mark Paluch
  * @author Jens Schauder
+ * @author Christoph Strobl
  */
 public interface PersistentProperty<P extends PersistentProperty<P>> {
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/springframework/data/mapping/PersistentProperty.java
```diff
@@ -276,6 +281,24 @@ default Field getRequiredField() {
...
+		return isEntity() && findAnnotation(Embedded.class) != null;
+	}
+
+	/**
+	 * @return {@literal true} if the property generally allows {@literal null} values;
+	 * @since 2.5
+	 */
+	default boolean isNullable() {
+
+		return (isEmbedded() && findAnnotation(Embedded.class).onEmpty().equals(OnEmpty.USE_NULL))
```
schauder, line 298: When a property is not embedded, it is never nullable. If that is intended, it should be described in the documentation and the superfluous parens around the first `&&` term should be removed. 

Or the expression should get fixed. Also we should have tests for the not embedded case.

schauder (changes requested):

dcominottim:
It'd be really nice if we could further customize this annotation with more parameters or pair it with another annotation to achieve the same effect from `@Embedded` +` @AttributeOverride` from `javax.persistence`. This way, the name of the inline fields could be customized and the embeddable objects could be re-used in different parent objects that need different field names for inline fields.

mp911de:
Embedded objects are conceptually stable regarding their property to field/column/… names. `@Embedded` allows you to provide a prefix to customize physical naming. An example use-case:

```java
class Address {
  String street, city;
}

class Order {

  @Embedded(prefix = "shipping_")
  Address shipping;

  @Embedded(prefix = "billing_")
  Address billing;
}
```

Rendering fields `shipping_street`, `shipping_city`, `billing_street`, and so on.

@dcominottim care to elaborate how else you'd like to customize field names?

dcominottim:
> Embedded objects are conceptually stable regarding their property to field/column/… names. `@Embedded` allows you to provide a prefix to customize physical naming. An example use-case:
> 
> ```java
> class Address {
>   String street, city;
> }
> 
> class Order {
> 
>   @Embedded(prefix = "shipping_")
>   Address shipping;
> 
>   @Embedded(prefix = "billing_")
>   Address billing;
> }
> ```
> 
> Rendering fields `shipping_street`, `shipping_city`, `billing_street`, and so on.
> 
> @dcominottim care to elaborate how else you'd like to customize field names?

Thanks for the reply, @mp911de. I had overlooked the 'prefix' param. It seems to work fine for embedded objects that aggregate multiple properties (such as your `Address` example). I think it should often work for value objects that encapsulate a single property as well (e.g.: `class CompanyName { String companyName; }`), although it'd still be useful for the parent to be able to fully override that -- in JPA, it'd be

```java
class Company {
  @Embedded
  @AttributeOverride(
    name = "companyName", // name of the target attribute
    column = @Column(name = "CMP_NAME") // name that will be used for persistence
  )
  CompanyName name;
}
```

Think of a scenario in which the company you work for has complex rules for data modeling and the data governance team may request you to use specific column/attribute names for certain tables/documents -- and they don't play well with Java/Spring Data attribute naming conventions. I am in such scenario right now and need to use @AttributeOverride in JPA projects.

On a separate note, does Spring Data fully support underscore in attribute names nowadays (as per your prefix example)? It didn't use to. Sorry if too obvious. https://github.com/spring-projects/spring-data-commons/pull/126

sseep:
Thank you for open source contribution.  The feature still be watched?

mp911de:
Superseded by #2294.

Issue #2230, opened 2020-12-21
Javadoc issue with CrudRepository#deleteAll() [DATACMNS-1848]


spring-projects-issues:
**[Michal Domagala](https://jira.spring.io/secure/ViewProfile.jspa?name=michaldo)** opened **[DATACMNS-1848](https://jira.spring.io/browse/DATACMNS-1848?redirect=false)** and commented

Javadoc for https://github.com/spring-projects/spring-data-commons/blob/master/src/main/java/org/springframework/data/repository/CrudRepository.java#L133
states:

```java
/**
  * Deletes all entities managed by the repository.
*/
void deleteAll();
```

But implementations, for example SimpleJpaRepository, delete all records from table, not just entities managed by persistence context

```java
@Override
@Transactional
public void deleteAll() {

  for (T element : findAll()) {
    delete(element);
  }
}
```

My question is: are Javadoc wrong or implementations are wrong

---
No further details from [DATACMNS-1848](https://jira.spring.io/browse/DATACMNS-1848?redirect=false)

mp911de:
Closing as per comment in https://github.com/spring-projects/spring-data-commons/pull/2265#issuecomment-753777238.

PR #2265, opened 2020-12-30
Clarify javadoc for deleteAll()

michaldo:
Word `managed` in JPA world has concrete meaning. The word should be avoided when not refers to entity managed by persistence context

This PR solves https://github.com/spring-projects/spring-data-commons/issues/2230

### DIFF
--- a/src/main/java/org/springframework/data/repository/CrudRepository.java
+++ b/src/main/java/org/springframework/data/repository/CrudRepository.java
@@ -128,7 +128,7 @@ public interface CrudRepository<T, ID> extends Repository<T, ID> {
 	void deleteAll(Iterable<? extends T> entities);
 
 	/**
-	 * Deletes all entities managed by the repository.
+	 * Deletes all entities from the table.
 	 */
 	void deleteAll();
 }


mp911de:
`CrudReposistory` isn't JPA-specific, it's a Spring Data-specific abstraction that is used with various data stores (MongoDB, Redis, Neo4j). All of them have different terminologies hence the current documentation is correct.

michaldo:
What do you think about change
>  Deletes all entities managed by the repository.

to 
>  Deletes all entities.

?

What information provides  `managed by the repository`? Only the method `deleteAll()` use the phrase "managed by the repository". What about other methods: are they works also over unmanaged entities or not?

I checked all module javadoc. Word "managed" is rather applied to types, not to instances (entities).

I think that "managed by the repository" is noise not information, misleading, inconsistent and should be removed

schauder:
It is certainly not noise. Deleting all entities would mean that calling `deleteAll` on a `PersonRepository` would also delete cars.

michaldo:
I just tried to avoid terminology collision. 
On one hand is JPA. Spring Data JPA is implementation, one of many, but significant: 140K hits on SO vs 85K, 73K, 47K for MongoDB, Redis, Neo4j respectively.
JPA sets concrete meaning for generic word "managed": there is difference between managed and detached entities.

On other hand abstraction Spring Data use general word commonly phrase "managed type" but  method "deleteAll()" use phrase "managed entity".

There are 4 clues that current comment is a noise:

1. However Spring Data is abstraction, it should avoid phrases used by implementation. It is not possible always, but in this case it is possible
2. Javadoc should be consistent and should not flip between "managed type" and "managed entity"
3. There is rule that something defined on higher level should not be repeated on lower level. In current case  higher level comment (class) says
*"Interface for generic CRUD operations on a repository for a specific type."* and lower level comment (method) should not repeat that (unless method javadac **changes** class javadoc)
4. Expectation that developers may think that PersonRepository#deleteAll will also also delete cars underestimates their perspicacity

PR #2285, opened 2021-01-27
Update auditing documentation.

christophstrobl:
Mention that even when only using the `@CreatedDate` & `@LastModifiedDate` annotations it is mandatory to enable auditing.
Add sample of using auditing metadata within an embedded entity.


mp911de:

PR #2286, opened 2021-01-29
Deprecate ChainedTransactionManager

### LINKED ISSUE #2232: Deprecate ChainedTransactionManager [DATACMNS-1817]
**[Mark Paluch](https://jira.spring.io/secure/ViewProfile.jspa?name=mp911de)** opened **[DATACMNS-1817](https://jira.spring.io/browse/DATACMNS-1817?redirect=false)** and commented

`ChainedTransactionManager` is the primary class in `org.springframework.data.transaction` that is used for multi-transactionmanager arrangements. It is useful to coordinate transactions across multiple resources especially when one transaction manager is a Spring one and the other is a foreign one.

Over time, we found that we're basically emulating distributed transactions on a best-effort basis with known gaps. These gaps lead to inconsistencies or unexpected behavior during rollbacks. Part of the problem is that resource/synchronization storage is a singleton `ThreadLocal`. Using two transaction managers based on `AbstractPlatformTransactionManager` causes the first transaction manager to handle all synchronizations regardless of their resource origin (primary and secondary transactional resources). If the second transaction manager commit fails, then already all synchronizations are processed and there's no way to recover.

Therefore, we're going to deprecate `ChainedTransactionManager` and the entire `org.springframework.data.transaction` that hosts support classes for multi-transactions

---

**Issue Links:**
- [DATACMNS-370](https://jira.spring.io/browse/DATACMNS-370) ChainedTransactionManager should only control transaction synchronization if all delegate transaction managers are set to NOT control synchronization
 (_**"supersedes"**_)
- [DATACMNS-1221](https://jira.spring.io/browse/DATACMNS-1221) ChainedTransactionManager - Issue with HibernateTransactionManager and KafkaTransactionManager
 (_**"supersedes"**_)
- [DATACMNS-1253](https://jira.spring.io/browse/DATACMNS-1253) ChainedTransactionManager and Ignite SpringTransactionManager
 (_**"supersedes"**_)
- [DATACMNS-1314](https://jira.spring.io/browse/DATACMNS-1314) ChainedTransactionManager does not handle transaction sychroriazation correctly
 (_**"supersedes"**_)
- [DATACMNS-1789](https://jira.spring.io/browse/DATACMNS-1789) Reactive ChainedTransactionManager
 (_**"supersedes"**_)
- [DATACMNS-1604](https://jira.spring.io/browse/DATACMNS-1604) Support 'afterAllCommit' in ChainedTransactionManager
 (_**"supersedes"**_)

mp911de:
We recommend to not use `ChainedTransactionManager` as it emulates distributed transactions without providing consistency guarantees that can end up in partially committed transactions. To attach to commit/rollback operations we recommend registering `TransactionSynchronization` instead.

Closes #2232.

### DIFF
--- a/src/main/java/org/springframework/data/transaction/ChainedTransactionManager.java
+++ b/src/main/java/org/springframework/data/transaction/ChainedTransactionManager.java
@@ -44,14 +44,32 @@ import org.springframework.util.Assert;
  * which means the {@link PlatformTransactionManager} most likely to break the transaction should be the <em>last</em>
  * in the list configured. A {@link PlatformTransactionManager} throwing an exception during commit will automatically
  * cause the remaining transaction managers to roll back instead of committing.
+ * <p />
+ * As consequence, a transaction can get into a state, where the first {@link PlatformTransactionManager} has committed
+ * its transaction and a subsequent {@link PlatformTransactionManager} failed to commit its transaction (e.g. caused by
+ * an I/O error or the transactional resource failed to commit for other reasons). In that case,
+ * {@link #commit(TransactionStatus)} throws a {@link HeuristicCompletionException} to indicate a partially committed
+ * transaction. Rollback isn't affected as the natural consequence of a missing commit is a rollback of a transactional
+ * resource. {@link ChainedTransactionManager} should be only used if the application can tolerate or recover from
+ * inconsistent state caused by partially committed transactions. In any other case, the use of
+ * {@link ChainedTransactionManager} is not recommended.
+ * <p/>
+ * Instead of using {@link ChainedTransactionManager} for attaching callbacks to transaction commit (pre commit/post
+ * commit), either register a {@link org.springframework.transaction.reactive.TransactionSynchronization} to explicitly
+ * follow transaction cleanup with simplified semantics in case of exceptions.
  *
  * @author Michael Hunger
  * @author Oliver Gierke
+ * @author Mark Paluch
  * @since 1.6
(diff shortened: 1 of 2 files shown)


schauder (approved):

src/main/java/org/springframework/data/transaction/ChainedTransactionManager.java
```diff
@@ -44,14 +44,32 @@
...
+ * an I/O error or the transactional resource failed to commit for other reasons). In that case,
+ * {@link #commit(TransactionStatus)} throws a {@link HeuristicCompletionException} to indicate a partially committed
+ * transaction. Rollback isn't affected as the natural consequence of a missing commit is a rollback of a transactional
+ * resource. {@link ChainedTransactionManager} should be only used if the application can tolerate or recover from
+ * inconsistent state caused by partially committed transactions. In any other case, the use of
+ * {@link ChainedTransactionManager} is not recommended.
+ * <p/>
+ * Instead of using {@link ChainedTransactionManager} for attaching callbacks to transaction commit (pre commit/post
+ * commit), either register a {@link org.springframework.transaction.reactive.TransactionSynchronization} to explicitly
+ * follow transaction cleanup with simplified semantics in case of exceptions.
```
OrangeDog, line 59: This "either" seems to be missing an "or".

OrangeDog:
This didn't make it into the release notes?
https://spring.io/blog/2021/04/20/what-s-new-in-spring-data-2021-0

PR #2294, opened 2021-02-04
Use contextual information when creating PersistentPropertyPaths.

### LINKED ISSUE #2293: PersistentPropertyPathFactory should use contextual property information.
In oder to preserve contextual information the `PersistentPropertyPathFactory` should obtain `EntityInformation` for properties from the `MappingContext` via their `PersistentProperty` representation instead of plain `TypeInformation` as the former contains more information about the actual type and signatures.

Related to: spring-projects/spring-data-mongodb#2803

christophstrobl:
In oder to preserve contextual information the `PersistentPropertyPathFactory` now obtains `EntityInformation` for properties from the `MappingContext` via their `PersistentProperty` representation instead of plain the `TypeInformation` as the former contains more information about the actual type and signatures.

Closes #2293

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.5.0-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


mp911de:
That's merged now.

PR #2299, opened 2021-02-16
Use NativeDetector.inNativeImage() instead of manual property check.

christophstrobl:
Closes: #2298
Relates to: spring-projects/spring-framework#25795


mp911de:

Issue #2313, opened 2021-02-25
TypeDiscoverer.getParameterTypes does not capture enclosing type parameter of non static inner class


christophstrobl:
```java
static class SomeGeneric<T> {

	private final T value;

	protected SomeGeneric(T value) {
		this.value = value;
	}
}

class NonStaticInnerWithGenericFieldInSuper<T> extends SomeGeneric<T> {

	public NonStaticInnerWithGenericFieldInSuper(T value) {
		super(value);
	}
}

@Test
void capturesEnclosingTypeParameterOfNonStaticInnerClass() {

	assertThat(PreferredConstructorDiscoverer.discover(NonStaticInnerWithGenericFieldInSuper.class)).satisfies(ctor -> {

		assertThat(ctor.getParameters()).hasSize(2);
		assertThat(ctor.getParameters().get(0).getName()).isEqualTo("this$0");
		assertThat(ctor.getParameters().get(1).getName()).isEqualTo("value");
	});
}
```

It seems `Constructor.getGenericParameterTypes()` does only return the enclosing types one. Using `Constructor.getParameters()` can help read the `Type` via `Parameter.getParameterizedType` of each `Parameter`.

PR #2314, opened 2021-02-25
Collect constructor type information via Constructor.getParameters()

christophstrobl:
This PR fixes an issue where we fail to detect all type arguments from a given `Constructor`. 
As calling `getGenericParameterTypes()` in some cases does not include all `Type`s, we now explicitly iterate over the parameters and extract the parameterized type that is used for creating the `TypeInformation`.

Closes: #2313

### DIFF
--- a/src/main/java/org/springframework/data/util/TypeDiscoverer.java
+++ b/src/main/java/org/springframework/data/util/TypeDiscoverer.java
@@ -20,6 +20,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/util/TypeDiscoverer.java
```diff
@@ -175,14 +176,10 @@ protected TypeDiscoverer(Type type, Map<TypeVariable<?>, Type> typeVariableMap)
...
 
-		Type[] types = constructor.getGenericParameterTypes();
-		List<TypeInformation<?>> result = new ArrayList<>(types.length);
-
-		for (Type parameterType : types) {
-			result.add(createInfo(parameterType));
-		}
-
-		return result;
+		return Arrays.stream(constructor.getParameters())
```
mp911de, line 179: Can we stick with `for` loops and the `ArrayList` using pre-initialized sizes?

christophstrobl: Replaced `stream()` with `for` loop in 53917d9

mp911de:

PR #2316, opened 2021-02-26
Support for jMolecules' Association type.

### LINKED ISSUE #2315: Support for jMolecules' Association type
It would be cool if Spring Data's object mapping infrastructure supported jMolecules `Association` type discovering it as logical association and automatically converting it into its identifier type and the underlying primitive value respectively:

```java
class Aggregate {
  Association<OtherAggregate, OtherAggregateIdentifier> association;
}

class OtherAggregateIdentifier implements Identifier {
  UUID id;
}
```

The former needs `PersistentProperty.isAssociation()` to be made aware of the type. The latter can be implemented by adding [the `Converter` implementations](https://github.com/xmolecules/jmolecules-integrations/tree/main/jmolecules-spring/src/main/java/org/jmolecules/spring) available in jMolecules `jmolecules-spring` artifact.

odrotbohm:
We no recognize properties of type `org.jmolecules.ddd.types.Association` as associations in our `PersistentProperty` model.

Also, we now register JMolecules' `Converter` implementations for `Association` and `Identifier` in `CustomConversions` so that they can persisted like their embedded primitive value out of the box.

Fixes #2315.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -339,6 +339,13 @@
 			<version>0.1.4</version>
 			<scope>test</scope>
 		</dependency>
+		
+		<dependency>
+			<groupId>org.jmolecules.integrations</groupId>
+			<artifactId>jmolecules-spring</artifactId>
+			<version>${jmolecules-integration}</version>
+			<optional>true</optional>
+		</dependency>
 
 	</dependencies>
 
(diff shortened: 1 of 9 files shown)


mp911de:
That's merged now.

PR #2311, opened 2021-02-24
Add @JpaRepositoryCombination function. avoid duplicate inheritance i…

caomingjie-code:
…nterface

- [x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [ ] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [ ] You submit test cases (unit or integration tests) that back your changes.
- [x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

PR #2317, opened 2021-03-01
I think this is a custom warehouse interface optimization problem

caomingjie-code:


caomingjie-code:
Custom repository interfaces cannot inherit more, for example:

A<T,ID> extends JpaRepersitory<T,ID> , CustomRepoersitory<T,ID>{},B extends A<Book,Integer>{}

In this case, an error will be reported. We should enhance the lookup function of the custom warehouse interface. Otherwise, the custom warehouse interface function is very weak. I think the custom warehouse interface function will be very flexible after enhancement.

mp911de:
This is a duplicate of #2311. We do not plan to add this contribution.

PR #2339, opened 2021-03-23
Document effect of using `AccessType(PROPERTY)` in combination with `Persistable`

### LINKED ISSUE #2338: Unknown column 'new' in 'field list' when implementing Persistable in a class with @AccessType(Type.PROPERTY)
My Spring Data JDBC entity is implementing `Persistable` and is annotated with `@AccessType(Type.PROPERTY)` at the class level.
When I invoke `save` on my repository, I get the following exception from the underlying JDBC implementation:

`java.sql.SQLException: Unknown column 'new' in 'field list'`

Indeed, looking at the executed query, Spring Data JDBC is also specifying the column `new` in the insert statement. I think it comes from the implementation of `org.springframework.data.domain.Persistable.isNew()`.

I take the opportunity to ask a question: is it true that implementing `Persistable` allows me to avoid to mark the primary key field with `@Id`, being the implementation of `org.springframework.data.domain.Persistable.getId()` enough for the framework to detect the primary key? It's not very clear from the documentation.

mp911de:
Closes #2338

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.5.0-SNAPSHOT</version>
+	<version>2.5.0-GH-2338-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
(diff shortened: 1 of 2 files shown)


schauder:
That's merged and ported back.

PR #2342, opened 2021-03-25
Improve error message if store module doesn't support a well-known fragment interface

### LINKED ISSUE #2341: Improve error messages if a repository cannot be created
**[Tera](https://jira.spring.io/secure/ViewProfile.jspa?name=JIRAUSER49634)** opened **[DATACMNS-1812](https://jira.spring.io/browse/DATACMNS-1812?redirect=false)** and commented

I'm trying to use Spring data JDBC with Query DSL in Kotlin. This is my repository

 

```java
@Repository
interface AgentRepository : PagingAndSortingRepository<Agent, Int>, QuerydslPredicateExecutor<Agent> {}
```

I've integrated these dependencies in my Gradle build:

```java
implementation 'org.springframework.boot:spring-boot-starter-data-jdbc:2.3.3'
implementation("com.querydsl:querydsl-core:4.4.0") 
kapt "com.querydsl:querydsl-apt:4.4.0:jpa"
```

As soon as I try to start my application I get this error:

```java
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'agentRepository' defined in it.dove.braniac.repositories.AgentRepository defined in @EnableJdbcRepositories declared on BraniacApplication: Invocation of init method failed; nested exception is org.springframework.data.repository.query.QueryCreationException: Could not create query for public abstract long org.springframework.data.querydsl.QuerydslPredicateExecutor.count(com.querydsl.core.types.Predicate)! Reason: No property count found for type Agent!
```

I can't understand what I'm doing wrong and if there is an incompatibility between Spring Data JDBC and QueryDSL

 

The same issue happens if I try to change from PagingAndSortingRepository to simple CrudRepository

---

**Affects:** 2.3.4 (Neumann SR4)

mp911de:
We now throw a `RepositoryCreationException` (or subclass) when a repository cannot be created due to a missing fragment, a fragment without implementation or if a well-known fragment is not supported by the repository factory.

Closes #2341

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.5.0-SNAPSHOT</version>
+	<version>2.5.0-GH-2341-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
@@ -339,7 +339,7 @@
 			<version>0.1.4</version>
 			<scope>test</scope>
 		</dependency>
-		
+
 		<dependency>
 			<groupId>org.jmolecules.integrations</groupId>
 			<artifactId>jmolecules-spring</artifactId>
--- a/src/main/java/org/springframework/data/repository/core/RepositoryCreationException.java
+++ b/src/main/java/org/springframework/data/repository/core/RepositoryCreationException.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.repository.core;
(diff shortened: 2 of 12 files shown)


odrotbohm:
What's the exact difference between `FragmentNotImplementedException` and `MissingFragmentException`? Superficially they sound very similar and I don't quite get the difference from the Javadoc

mp911de:
`MissingFragmentException` is used when we identify custom methods and do not find an implementation for the method to be called. `MFE` isn't tied to a particular fragment, it's expressing that the composition doesn't contain what we're looking for. `FragmentNotImplementedException` signals that we identified a fragment but it has no implementation attached. `FragmentNotImplementedException` can only happen when configuring `RepositoryComposition` manually and it should not occur when bootstrapping repositories using repository scanning.

odrotbohm:
Does `IncompleteFragmentException` describe the situation better than `MissingFragmentException` then? A fragment is the entire interface, right, not just an individual method within it?

mp911de:
A fragment is an entire interface. I think `IncompleteRepositoryCompositionException` puts it better because we're in the context of a `RepositoryComposition` and we can't tell in that case which fragment is affected.

odrotbohm:
Any reason for the move to `RepositoryException` and drop the `Creation`? That sounds rather generic. Isn't all that structural validation tied to the repository creation?

mp911de:
Most exceptions happen indeed during repository creation. Some can happen when calling a repository method for a composition that wasn't configured properly. I'm not sure we can catch all failures early on. If we roll back runtime (invocation checks) for missing fragments/fragment implementations during repository method invocation, we could retain the concept of a creation exception.

odrotbohm:
How about `RepositorySetupException`? That slightly nudges into the direction of the arrangement of the repository.

mp911de:
That's merged now.

PR #2344, opened 2021-03-31
Support for association target type detection for jMolecules.

odrotbohm:
We now detect the component type of jMolecules' `Association` as association target type in `AbstractPersistentProperty`. `AnnotationBasedPersistentProperty` keeps the lookup of the explicit `@Reference` around but falls back to the `Association` analysis for that.

It was also changed to return `null` in case no association target type could be looked up instead of returning the sole property's type in case that is an entity. This was done to satisfy the behavior documented on the interface but also to avoid the call to `isEntity()` during the calculation which might use association information in turn and thus lead to a stack overflow.


odrotbohm:
Updates done, ready for review.

PR #2364, opened 2021-04-26
Public constructor preference in PreferredConstructorDiscoverer.Discoverers.DEFAULT

### LINKED ISSUE #2363: In case of multiple constructor candidates give preference to the public one
Speaking of `immutable dto classes`, there are many cases where there are several non-default constructors in it. Current implementation of PreferredConstructorDiscoverer.Discoverers.DEFAULT can't resolve these cases without `@PersistenceConstructor`. 
Usually this is not a big problem, however when using libraries such as [Lombok](https://projectlombok.org/) or [Immutables](https://immutables.github.io/) this adds additional inconvenience. These libraries can generate additional private constructors for builders `contructor(builder)` or copy methods `copyOf(self)`/`with*`

The idea is not to return null when several non-default constructors are found, but try to find one public among them (if it is the only public one) and return it.

AbstractCoderX:
[resolves #2363 ]

### DIFF
--- a/src/main/java/org/springframework/data/mapping/model/PreferredConstructorDiscoverer.java
+++ b/src/main/java/org/springframework/data/mapping/model/PreferredConstructorDiscoverer.java
@@ -22,6 +22,7 @@ import kotlin.reflect.jvm.ReflectJvmMapping;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Modifier;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


mp911de:
Closing this one as per https://github.com/spring-projects/spring-data-commons/issues/2363#issuecomment-831130006.

Since Immutables style customization affects how classes are generated, it makes sense to find tweaks within Immutables (e.g. adding a particular annotation to each constructor) as we cannot make assumptions over the actual customization possibilities in Immutables.

PR #2367, opened 2021-05-03
Remove `PersistentEntity` from `MappingContext` upon mapping metadata initialization errors

gregturn:


gregturn:

src/main/java/org/springframework/data/mapping/context/AbstractMappingContext.java
```diff
@@ -378,25 +379,21 @@ public E getPersistentEntity(P persistentProperty) {
...
-			} catch (RuntimeException e) {
-				persistentEntities.remove(typeInformation);
-				throw e;
+			if (persistentPropertyAccessorFactory.isSupported(entity)) {
+				entity.setPersistentPropertyAccessorFactory(persistentPropertyAccessorFactory);
 			}
 
 		} catch (BeansException e) {
 			throw new MappingException(e.getMessage(), e);
+		} catch (RuntimeException e) {
```
mp911de, line 394: We should also remove the entity from the cache upon `BeansException`

mp911de:

Issue #2103, opened 2020-04-15
Sort.and() does not handle TypedSort without explicit order  [DATACMNS-1704]


spring-projects-issues:
**[hgarus](https://jira.spring.io/secure/ViewProfile.jspa?name=hgarus)** opened **[DATACMNS-1704](https://jira.spring.io/browse/DATACMNS-1704?redirect=false)** and commented

This might me misunderstanding the API. I did try to sort by multiple properties doing the following:

```java
Sort<MyEntity> myEntity = Sort.sort(MyEntity.class).by(MyEntity::getProperty1)
    .and(Sort.sort(MyEntity.class).by(MyEntity::getProperty2))
```

The result is a Sort by property2, looking at the code the reason seems to be Sort.and() accessing Sort.orders directly, which is empty for TypedSort

---
No further details from [DATACMNS-1704](https://jira.spring.io/browse/DATACMNS-1704?redirect=false)

PR #2377, opened 2021-05-15
Address the issue of missing sort criteria in the case of Entity class (TypedSort)

HiuKwok:
Closes https://github.com/spring-projects/spring-data-commons/issues/2103
This is a Pull Request to address the issue reported on DATACMNS-1704, to add a getter method for property `orders` on both classes `Sort` and its child class `TypedSort` in order to handle the case that the user invokes multiple sort orders with custom entity class.


mp911de:
Thanks for your pull request. We can fix the issue with less code changes. Instead of exposing `getOrders`, we can call `this.toList()`. Given the huge number of commits, we're going to pick your test only and apply a fix from our side.

PR #2382, opened 2021-06-16
Allow for a module specific inclusion for DTO projections.

schauder:
See spring-projects/spring-data-jpa#2009


mp911de:
That's merged, polished and backported now.

PR #2389, opened 2021-06-23
Implement default Persistable::isNew

quaff:
Make subclasses concise from boilerplate code

### DIFF
--- a/src/main/java/org/springframework/data/domain/Persistable.java
+++ b/src/main/java/org/springframework/data/domain/Persistable.java
@@ -45,5 +45,7 @@ public interface Persistable<ID> {
 	 *
 	 * @return if {@literal true} the object is new.
 	 */
-	boolean isNew();
+	default boolean isNew() {
+		return getId() == null;
+	}
 }


mp911de:
Thanks for your suggestion. We want to keep the current arrangement as it requires consideration on the implementing side how to design the `is-new` check properly. In the case of optimistic locking, a simple non-null check isn't sufficient. Also, when working with primitive identifiers, the Id value can't be ever `null` so the implementation would be misleading.

That being said, we're closing this pull request without merging it.

odrotbohm:
I wonder what exactly the use case is for implementing `Persistable` but defaulting `isNew()` to the identifier null check. After all, this is exactly what you get if you *don't* implement `Persistable`. I.e. you'd only implement that interface if you *don't* want that default to apply, right?

quaff:
> Thanks for your suggestion. We want to keep the current arrangement as it requires consideration on the implementing side how to design the `is-new` check properly. In the case of optimistic locking, a simple non-null check isn't sufficient. Also, when working with primitive identifiers, the Id value can't be ever `null` so the implementation would be misleading.
> 
> That being said, we're closing this pull request without merging it.

Primitive identifiers are not allowed for `Persistable`, you can't implements `Persistable<long>`.

PR #2391, opened 2021-06-24
Fix wrong example code

quaff:
1. Add missing `s` in `@RequiredArgsConstructor`
2. Add `final` modifier to mark field as required


mp911de:
Good catch. Thank you for your contribution. That's merged, polished, and backported now.

PR #2394, opened 2021-06-25
Don't create PersistentEntity for Nullable wrapper type.

christophstrobl:
Detect a _nullable_ wrapper type (like `java.util.Optional`) and avoid creating an entity for it. 
Also updates entity detection for properties that nest the target type deep within the generic structure as well as map key types.

delivers-input-to: spring-projects/spring-data-mongodb#3656

Tested against Modules for: JPA, MongoDB, Redis


mp911de:

PR #2397, opened 2021-06-29
[Documentation][Fix] Streamable example using RequiredArgConstructor

cailliaud:
Example of Streamable using @RequiredArgConstructor(staticName = "of") won't generate a static "of" waiting a Streamable if the field is not final or at least with @lombok.NonNull.
Based on Lombok documentation : https://projectlombok.org/features/constructor

Without a static constructor "of" containing the Streamable it will throw an : 
```
java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class [...] Products
```


mp911de:
Thanks for your contribution. That was already fixed with #2391

PR #2399, opened 2021-07-05
Polish

snicoll:
A `public` bean method that returns a private type is unusual and prevents the processing of such bean in reflection-free use cases. There was also an unnecessary qualifier in the implementation.


mp911de:

PR #2402, opened 2021-07-05
Corrects the description of constructor detection.

### LINKED ISSUE #2401: Correct documentation how a constructor is chosen.

schauder:
Closes #2401

### DIFF
--- a/src/main/asciidoc/object-mapping.adoc
+++ b/src/main/asciidoc/object-mapping.adoc
@@ -17,10 +17,10 @@ This means we need two fundamental steps:
 Spring Data automatically tries to detect a persistent entity's constructor to be used to materialize objects of that type.
 The resolution algorithm works as follows:
 
-1. If there's a no-argument constructor, it will be used.
+1. If there is a single constructor, it will be used.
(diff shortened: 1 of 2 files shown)


src/main/asciidoc/object-mapping.adoc
```diff
@@ -17,10 +17,10 @@ This means we need two fundamental steps:
 Spring Data automatically tries to detect a persistent entity's constructor to be used to materialize objects of that type.
 The resolution algorithm works as follows:
 
-1. If there's a no-argument constructor, it will be used.
+1. If there is a single constructor, it will be used.
```
Buzzardo, line 20: In all three lines, change "will be" to "is"

---------

This is a general rule. The idea is to keep the tense consistent, so we use the indefinite present (also called the simple present) tense.

Buzzardo (changes requested):

src/main/asciidoc/object-mapping.adoc
schauder: Thanks for the explanation. I think the use of "will be" is a wrong translation of passive voice which is required German.

mp911de:

PR #2403, opened 2021-07-06
Accept `JsonProvider` upon `JsonProjectingMethodInterceptorFactory` creation

sokomishalov:
It causes NCDFE in the case of excluding `json-smart` from the classpath.
While calling `Configuration::build` without providing `JsonProvider` it takes the default one (`JsonSmartJsonProvider`) with bad consequences for spring-boot context initialization (like [that](https://github.com/zalando/logbook/issues/1074))


mp911de:

PR #2406, opened 2021-07-07
`Repositories` now allows lookup of parent repositories for sub-types.

odrotbohm:
When inheritance is used for aggregates, lookups of the repository for a Child aggregate instance have so far failed to return a repository registered for the Parent. Client could had to consider that scenario explicitly.

This commit introduces an additional lookup step in case we cannot find a repository for an aggregate type immediately. In this case, we then check for assignability of any of the known aggregate types we have registered repositories for and the type requested. I.e. for a request for the repository of Child, a repository, explicitly registered to manage Child instances would still be used. In the sole presence of a repository managing Parent instances, that would be returned for the request for Child, too.


odrotbohm:
I think it's safe to back-port to 2021.0. Nothing particular preventing it from even going into 2020.0, but would understand if you'd like to minimize the risk for that branch.

mp911de:

Issue #2409, opened 2021-07-09
PersistentEntity type detection for PersistentProperties should consider association target type


odrotbohm:
The entity type detection currently doesn't consider the association target type in the first place. In scenarios, in which the actually declared property type is leads to a different entity consideration, this might lead to wrong inspection results.

Issue #2408, opened 2021-07-09
Consolidate access to `TypeInformation<?>` VS. `Class<?>` on `PersistentProperty`


odrotbohm:
Currently, we have a few inconsistencies in naming methods that return `Class<?>` VS. `TypeInformation<?>`. We should name the ones returning the latter `…TypeInformation()` consistently.

PR #2410, opened 2021-07-09
Fixes for #2408 and #2409.

odrotbohm:
See the commit messages for details. Needs spring-projects/spring-data-jpa#2255 and spring-projects/spring-data-mongodb#3701 merged, too.

### DIFF
--- a/src/main/java/org/springframework/data/mapping/PersistentProperty.java
+++ b/src/main/java/org/springframework/data/mapping/PersistentProperty.java
@@ -70,8 +70,20 @@ public interface PersistentProperty<P extends PersistentProperty<P>> {
 	 * {@link Map}'s value type transparently.
 	 *
 	 * @return
+	 * @deprecated since 2.6 for removal in 3.0. Use {@link #getPersistentEntityTypeInformation()} instead.
 	 */
+	@Deprecated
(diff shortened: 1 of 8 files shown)


src/main/java/org/springframework/data/mapping/PersistentProperty.java
```diff
@@ -70,9 +70,21 @@
...
+	/**
+	 * Returns the {@link TypeInformation} if the property references a {@link PersistentEntity}. Will return
+	 * {@literal null} in case it refers to a simple type. Will return {@link Collection}'s component type or the
+	 * {@link Map}'s value type transparently.
+	 *
+	 * @return
+	 * @since 2.6
+	 */
+	Iterable<? extends TypeInformation<?>> getPersistentEntityTypeInformation();
+
```
christophstrobl, line 87: The documentation is a bit confusing as it states the method to return `null` for a method thats return type is an `Iterable`. Wouldn't _empty_ be a better fit?

odrotbohm: Good catch. Indeed, the Javadoc still stems from the times that the original method which the comment was copied from returned a single type, not a collection.

PR #2416, opened 2021-07-17
Fix typo in entity state detection documentation

schlagi123:
- [x] You submit test cases (unit or integration tests) that back your changes. (asciidoc only)
- [x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only). (asciidoc only)

### DIFF
--- a/src/main/asciidoc/is-new-state-detection.adoc
+++ b/src/main/asciidoc/is-new-state-detection.adoc
@@ -7,14 +7,14 @@ The following table describes the strategies that Spring Data offers for detecti
 [options = "autowidth",cols="1,1"]
 |===
 |`@Id`-Property inspection (the default)
-|By default, Spring Data inspects the version property of the given entity.
-If the identifier property is `null` or `0` in case of primitive types, then the entity is assumed to be new.
+|By default, Spring Data inspects the id property of the given entity.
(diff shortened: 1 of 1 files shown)


src/main/asciidoc/is-new-state-detection.adoc
```diff
@@ -7,14 +7,14 @@ The following table describes the strategies that Spring Data offers for detecti
 [options = "autowidth",cols="1,1"]
 |===
 |`@Id`-Property inspection (the default)
-|By default, Spring Data inspects the version property of the given entity.
-If the identifier property is `null` or `0` in case of primitive types, then the entity is assumed to be new.
+|By default, Spring Data inspects the id property of the given entity.
+If the id property is `null` or `0` in case of primitive types, then the entity is assumed to be new.
```
mp911de, line 11: Any reason you're changing `identifier` into `id` in random places?

schlagi123: In some places in the text there is Id, in other places there is an identifier. I thought I would unify it in this section. I don't care whether it's id or identifier, I just thought id is closer to @Id than identifier.

mp911de:

PR #2420, opened 2021-07-26
Provide class for inspecting nested projections.

meistermeier:
As a result there will be a `Map<PropertyPath, Boolean>`, indicating if the property path "ends" in an entity or not, that can get used as input for a selection/mapping filter in the dedicated store modules.
Maybe it might makes sense to wrap the `Map<PropertyPath, Boolean>` into an own class. For us it is sufficient right now to use the map.


christophstrobl:
Thanks @meistermeier! Now available on _main_ now.

PR #2421, opened 2021-07-28
Add FluentQuery support to `QueryByExampleExecutor` and `ReactiveQueryByExampleExecutor`

### LINKED ISSUE #2228: Add support for QueryDSL and Query by Example projections in repositories
**[marcusvoltolim](https://jira.spring.io/secure/ViewProfile.jspa?name=JIRAUSER50351)** opened **[DATACMNS-1813](https://jira.spring.io/browse/DATACMNS-1813?redirect=false)** and commented

https://jira.spring.io/browse/DATAJPA-1796

---

**Affects:** 2.3.4 (Neumann SR4)

**Reference URL:** https://jira.spring.io/browse/DATAJPA-1796

mp911de:
`FluentQuery` allows extending a query specification initially defined by a Example probe or a Querydsl Predicate and fetching the actual result through a functional programming model:

```java
interface PersonRepository extends QuerydslPredicateExecutor<Person> {
  <S extends T, R> R findBy(Predicate predicate, Function<FluentQuery.FetchableFluentQuery<S>, R> queryFunction);
}

PersonRepository repo = …;

List<PersonProjection> result = repo.findBy(QPerson.person.name.eq("Walter"), q -> q.sort(Sort.by("lastname")).as(PersonProjection.class).all());

Stream<PersonProjection> result = repo.findBy(QPerson.person.name.eq("Walter"), FetchableFluentQuery::stream);
```

Closes #2228


src/main/java/org/springframework/data/querydsl/QuerydslPredicateExecutor.java
```diff
@@ -108,4 +110,15 @@
...
+	/**
+	 * Returns entities matching the given {@link Predicate} applying the {@link Function queryFunction} that defines the
+	 * query and its result type.
+	 *
+	 * @param predicate must not be {@literal null}.
+	 * @param queryFunction the query function defining projection, sorting, and the result type
+	 * @return all entities matching the given {@link Predicate}.
+	 * @since 2.6
+	 */
+	<S extends T, R> R findBy(Predicate predicate, Function<FluentQuery.FetchableFluentQuery<S>, R> queryFunction);
```
michael-simons, line 123: Why not defaulting it?

src/main/java/org/springframework/data/repository/query/QueryByExampleExecutor.java
```diff
@@ -86,4 +87,15 @@
...
+	/**
+	 * Returns entities matching the given {@link Example} applying the {@link Function queryFunction} that defines the
+	 * query and its result type.
+	 *
+	 * @param example must not be {@literal null}.
+	 * @param queryFunction the query function defining projection, sorting, and the result type
+	 * @return all entities matching the given {@link Example}.
+	 * @since 2.6
+	 */
+	<S extends T, R> R findBy(Example<S> example, Function<FluentQuery.FetchableFluentQuery<S>, R> queryFunction);
```
michael-simons, line 100: Same.

michael-simons (commented):
The addition in general seems nice, but I wonder why you don't default the methods and throw an `UnsupportedOperationException` ?

mp911de:
This extension provides much more flexibility and is a pre-requisite to turn the other methods into default ones. We do not want to raise false expectations in terms of supported API to then throw an exception if e.g. streaming or projections aren't supported. 

All default repository methods are invoked on the interface as we do not delicate these calls into the actual implementations.

michael-simons:
But the change here will break all other modules.
What if a module cant or wont implement them and adds just stub throwing an exception?

michael-simons:
Another option might be an additional `FluentQuerydslPredicateExecutor` ?

mp911de:
> But the change here will break all other modules.
> What if a module cant or wont implement them and adds just stub throwing an exception?

That's a consequence of the current arrangement. Modules can still throw an `UnsupportedOperationException` in that case.

> Another option might be an additional FluentQuerydslPredicateExecutor ?

Introducing another set of interfaces (for imperative/reactive and Querydsl/Query-by-Example) break the concise programming model as the fluent method cannot be discovered from the marker interface that introduced Querydsl/Query-by-Example functionality to a repository. Each module would have to consider additional marker interfaces to contribute further fragments. As a consequence, we end up with a much more complex arrangement.

michael-simons:
> That's a consequence of the current arrangement. Modules can still throw an `UnsupportedOperationException` in that case.

I don't understand this sentiment in connection with your original above:

> We do not want to raise false expectations in terms of supported API to then throw an exception if e.g. streaming or projections aren't supported.

So either commons or a module will throw an exception and leave unfulfilled expectations. Doesn't make much sense to me.

Note that I am totally in for this new feature, but breaking all modules with the reasoning here doesn't work for me.

michael-simons:
Please add a note to the relevant `FluentQuery.*` interfaces that the builders are immutable in the sense that each step returning a `FluentQuery.*` again returns a complete new instance and does not modify the target.

mp911de:
That's merged now.

PR #2456, opened 2021-09-15
Set repository FactoryBean.OBJECT_TYPE_ATTRIBUTE as Class.

christophstrobl:
We should set the repository type as type hint for the bean container if we can. We should introduce a control mechanism whether we can provide such a hint as the type hint triggers class loading. In some arrangements (e.g. EclipseLink with load-time weaving) we must defer class loading as much as possible.

See also https://github.com/spring-projects-experimental/spring-native/blob/main/spring-aot/src/main/java/org/springframework/data/RepositoryFactoryBeanPostProcessor.java

for further details.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>2.5.5-SNAPSHOT</version>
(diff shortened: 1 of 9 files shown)


mp911de:
No longer required (context: AOT/Native Object Factory hints).

christophstrobl:
obsolete

PR #2471, opened 2021-09-28
chore: add Smallrye Mutiny support

hantsy:
aligned to Spring 5.3.10 which added SmallRye Mutiny support, see: https://github.com/spring-projects/spring-framework/pull/27331

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -148,6 +148,14 @@
 			<optional>true</optional>
 		</dependency>
 
+		<!-- SmallRye Mutiny, see:https://smallrye.io/smallrye-mutiny/ -->
+		<dependency>
+			<groupId>io.smallrye.reactive</groupId>
+			<artifactId>mutiny</artifactId>
+			<version>${smallrye-mutiny}</version>
+			<optional>true</optional>
+		</dependency>
+
 		<!--  Querydsl -->
 
 		<dependency>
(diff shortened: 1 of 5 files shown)


hantsy:
depends on https://github.com/spring-projects/spring-data-build/pull/1545

mp911de:
Thanks a lot for your contribution. For the time being we would like to include only the addition of Mutiny to our registry and adapter infrastructure. Care to remove the repository base interfaces as these aren't required and we do not have native implementations using Mutiny?

hantsy:
Is there any Repository implementations using RxJava 2/3?  

I would like to add it firstly, and then add a standalone spring data hibernate module(active and blocking). Currently Spring Data JPA does not support reactive APIs, but HibernateReacitve provides **JPA like** APIs to reactive world),  Hibernate Reactive has built-in SmallRye Mutiny support.

For those RDBMS developers who are familiar with JPA, migrating to Hibernate Reactive is more reasonable than using R2dbc(not mature as Jdbc till now).

mp911de:
>  would like to add it firstly, and then add a standalone spring data hibernate module(active and blocking). 

For the time being, it can be a local interface in a Spring Data Hibernate Reactive module.

> Is there any Repository implementations using RxJava 2/3?

Kind of. With Spring Data 2.x we support RxJava 1 while there is no base RxJava 1 base repository. Generally, Spring Data creates adapters around reactive types so one can use e.g. MongoDB through RxJava 2/3. In the early days, Couchbase was built on RxJava 1 but the repository could be used through `ReactiveCrudRepository`.

hantsy:
I see, the ReactiveAdapter can convert between them,  for Spring projects, Reactor is enough.

hantsy:
@mp911de  Updated and rebased on main.

mp911de:

PR #2480, opened 2021-10-12
Fix Javadoc in QueryByExampleExecutor

alikilickaya:
## What
Fixing the Javadoc of `QueryByExampleExecutor.findOne(Example<S> example)`.

## Why
`QueryByExampleExecutor.findOne(Example<S> example)` doesn't return `null`  if none was found. It returns `Optional.empty()`.


mp911de:

PR #2490, opened 2021-10-25
Add PageCollectors into Util

zorglube:

### DIFF
--- a/src/main/java/org/springframework/data/mapping/PropertyPath.java
+++ b/src/main/java/org/springframework/data/mapping/PropertyPath.java
@@ -47,7 +47,6 @@ public class PropertyPath implements Streamable<PropertyPath> {
 	private static final String PARSE_DEPTH_EXCEEDED = "Trying to parse a path with depth greater than 1000! This has been disabled for security reasons to prevent parsing overflows.";
 
 	private static final String DELIMITERS = "_\\.";
-	private static final String ALL_UPPERCASE = "[A-Z0-9._$]+";
 	private static final Pattern SPLITTER = Pattern.compile("(?:[%s]?([%s]*?[^%s]+))".replaceAll("%s", DELIMITERS));
 	private static final Pattern SPLITTER_FOR_QUOTED = Pattern.compile("(?:[%s]?([%s]*?[^%s]+))".replaceAll("%s", "\\."));
 	private static final Pattern NESTED_PROPERTY_PATTERN = Pattern.compile("\\p{Lu}[\\p{Ll}\\p{Nd}]*$");
--- a/src/main/java/org/springframework/data/util/PageCollectors.java
+++ b/src/main/java/org/springframework/data/util/PageCollectors.java
@@ -21,7 +21,35 @@ import org.springframework.data.domain.PageImpl;
 import org.springframework.data.domain.Pageable;
 
 /**
- * Utility methods to work with {@link Page}s.
+ * Utility methods to work with {@link Stream} and {@link Page}s. </br>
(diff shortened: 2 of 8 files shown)


odrotbohm:
Is there any existing ticket that describes, what issue this proposal is trying to solve? If not, would you mind to elaborate?

zorglube:
> Is there any existing ticket that describes, what issue this proposal is trying to solve? If not, would you mind to elaborate?

As fa as I know their is no ticket related to this PR. 

The problem I tried to handle is just a small writing 'issue'. 
I you like to use `Spring-Data` to get a `Page<T>` of something an then appli any process to the Page before returning the processed Page you'll have to write something like : 
```java 
class Something {

  public Page getProcessedPage(){
    List<T> items = repo.getPage().stream().map(process).collect(toList());
    Page<T> page = new PageImpl(items, items.size(), ...)
  }

}
```

odrotbohm:
You can apply processing to elements of a`Page` via its `….map(…)` method. We do not offer any methods to filter or reorder the elements, as that would by definition result in invalid `Page`s regarding their metadata and position within the overall set of elements they provide a view into.

The code you show effective produces broken `Page` instances as the ordering potentially does not match the `Sort` provided in the `Pageable`, it will be considered the last page if the filtering of the elements drops elements etc. This is not idiomatic usage of a `Page` instance. I'd love to learn what the *actual* use case for the post-processing and retaining the original metadata is, but as it stands now, I'm rather inclined to reject this PR.

zorglube:
I forgotted another use cas I had : 
```java 
import static org.springframework.data.util.PageCollectors.toFilteredSortedPage;

class Something {
  
  private final Predicate<T> filter;
  private final Comparator<T> order;

  public Page getProcessedPage(Pageable pageable) {
    List<T> list_1 = repo.getSomeData(); 
    List<T> list_2 = repo/service/watever.getSomeData(); 
    return list_1.stream().map/filter/sort( ... `imply list_2` ... ).collect(toFilteredSortedPage(pageable, filter, order));
  }

}
```

zorglube:
> You can apply processing to elements of a`Page` via its `….map(…)` method. We do not offer any methods to filter or reorder the elements, as that would by definition result in invalid `Page`s regarding their metadata and position within the overall set of elements they provide a view into.

Yes I'm aware of the `Page<T>.map(...)`. 

> 
> The code you show effective produces broken `Page` instances as the ordering potentially does not match the `Sort` provided in the `Pageable`, it will be considered the last page if the filtering of the elements drops elements etc. This is not idiomatic usage of a `Page` instance. I'd love to learn what the _actual_ use case for the post-processing and retaining the original metadata is, but as it stands now, I'm rather inclined to reject this PR.

Can you wait a few, I'll dig to find my old use case, as you ca see it was four months ago. 
Beside that, four months ago, I was already aware of the `Page<T>.map(...)` and I didn't found a solution, that's why I wrote those `Collectors`.

odrotbohm:
Sure thing. Be advised thought that we consider everything `util` *internal* utilities. I.e. unless we actually need code in there ourselves, we're not going to add code there, especially if it's only convenience, as it imposes maintenance cost for no real benefit to the team.

zorglube:
> Sure thing. Be advised thought that we consider everything `util` _internal_ utilities. I.e. unless we actually need code in there ourselves, we're not going to add code there, especially if it's only convenience, as it imposes maintenance cost for no real benefit to the team.

Yes I understand ;-) 
I purposed it because I "needed" it, and I thought someone else could need it AND I though it might be good to provide it through Spring.

mp911de:
Closing as per the previous comments.

PR #2491, opened 2021-10-27
Clarify behavior of `PropertyPath`

schauder:


mp911de:

PR #2517, opened 2021-12-17
Fixed resolving generic types for vavr maps.

nexx512:
Currently the generic types of vavr maps are not detected correctly. This should be fixed with this PR.


nexx512:
Are there chances that this will be fixed soon? Right now we are stuck with Spring 2.5.x and can't update our project to newer Spring versions.

odrotbohm:
That's applied. See [my comment](https://github.com/spring-projects/spring-data-commons/pull/2511#issuecomment-1041585859) on #2511 for details.

PR #2523, opened 2022-01-05
Expose `RepositoryMetadata` domain and id types as `TypeInformation`

### LINKED ISSUE #2518: RepositoryMetadata does not support generics identifiers and domain types
If I have a class Id<T> and a domain class Domain<T> and I defined a repository as:
```java
public interface StringDomainRepository extends CrudRepository<Id<String>, Domain<String>> {}
```
Then the RepositoryMetadata getIdType() and getDomainType() methods will return the erased types (Id and Domain) even though the information would be there and could be returned as Type rather than Class<?>.

I propose to change the interface as:
```java
/**
 * Metadata for repository interfaces.
 *
 * @author Oliver Gierke
 * @author Mark Paluch
 */
public interface RepositoryMetadata {

	/**
	 * Returns the id class the given class is declared for.
	 *
	 * @return the id class of the entity managed by the repository.
	 */
	TypeInformation<?> getIdType();

	/**
	 * Returns the domain class the repository is declared for.
	 *
	 * @return the domain class the repository is handling.
	 */
	TypeInformation<?> getDomainType();

	/**
	 * Returns the repository interface.
	 *
	 * @return
	 */
	Class<?> getRepositoryInterface();

	/**
	 * Returns the type {@link Method} return type as it is declared in the repository. Considers suspended methods and
	 * does not unwrap component types but leaves those for further inspection.
	 *
	 * @param method
	 * @return
	 * @since 2.4
	 */
	TypeInformation<?> getReturnType(Method method);

	/**
	 * Returns the domain class returned by the given {@link Method}. In contrast to {@link #getReturnType(Method)}, this
	 * method extracts the type from {@link Collection}s and {@link org.springframework.data.domain.Page} as well.
	 *
	 * @param method
	 * @return
	 * @see #getReturnType(Method)
	 */
	Class<?> getReturnedDomainClass(Method method);

	/**
	 * Returns {@link CrudMethods} meta information for the repository.
	 *
	 * @return
	 */
	CrudMethods getCrudMethods();

	/**
	 * Returns whether the repository is a paging one.
	 *
	 * @return
	 */
	boolean isPagingRepository();

	/**
	 * Returns the set of types the repository shall be discoverable for when trying to look up a repository by domain
	 * type.
	 *
	 * @see Repositories#getRepositoryFor(Class)
	 * @return the set of types the repository shall be discoverable for when trying to look up a repository by domain
	 *         type, must not be {@literal null}.
	 * @since 1.11
	 */
	Set<Class<?> getAlternativeDomainTypes();

	/**
	 * Returns whether the repository is a reactive one, i.e. if it uses reactive types in one of its methods.
	 *
	 * @return
	 * @since 2.0
	 */
	boolean isReactiveRepository();
}
```
This way lower level components, like for example GenericConverters, would be able to have the full type information.

alienisty:
RepositoryMetadata now provides full type information for identifiers and domain types.

Mainly this change is for supporting fully typed type conversions where the converted instance does not convey the fully reified type information which is instead obtainable from the repository interface signature.

Closes #2518

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -259,12 +259,6 @@
 		</dependency>
 
 		<dependency>
-			<groupId>jakarta.xml.bind</groupId>
-			<artifactId>jakarta.xml.bind-api</artifactId>
-			<version>2.3.3</version>
-		</dependency>
-
-		<dependency>
 			<groupId>com.sun.xml.bind</groupId>
 			<artifactId>jaxb-impl</artifactId>
 			<version>2.3.5</version>
--- a/pom.xml
+++ b/pom.xml
@@ -259,6 +259,12 @@
 		</dependency>
 
 		<dependency>
+			<groupId>jakarta.xml.bind</groupId>
+			<artifactId>jakarta.xml.bind-api</artifactId>
+			<version>2.3.3</version>
+		</dependency>
+
+		<dependency>
 			<groupId>com.sun.xml.bind</groupId>
 			<artifactId>jaxb-impl</artifactId>
 			<version>2.3.5</version>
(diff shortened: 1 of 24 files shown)


mp911de:

PR #2538, opened 2022-01-24
Introduce List-based repository interfaces

### LINKED ISSUE #2535: Create additional List-based repository interfaces
`ListCrudRepository`
`ListQuerydslPredicateExecutor`
`ListQueryByExampleExecutor`

These interfaces should look like there existing counterparts, but return a `List` where the existing ones return an `Iterable`.

schauder:
This introduces the interfaces
* `ListCrudRepository`
* `ListQuerydslPredicateExecutor`
* `ListQueryByExampleExecutor`

They return `List` instead of `Iterable` when returning multiple entities.
They can be used in the same way the interfaces without the `List` prefix can be used.

Closes #2535.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>3.0.0-SNAPSHOT</version>
+	<version>3.0.0-2535-list-interfaces-SNAPSHOT</version>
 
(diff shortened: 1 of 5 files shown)


mp911de:
How do we want to go about `PagingAndSortingRepository`? Do we want to add the `List`-based variant after decoupling `PagingAndSortingRepository` from `CrudRepository`?

schauder:
That's my plan.

Issue #1575, opened 2017-08-01
Support Reactive Repositories in Jackson2RepositoryPopulatorFactoryBean [DATACMNS-1133]


spring-projects-issues:
**[Juergen Zimmermann](https://jira.spring.io/secure/ViewProfile.jspa?name=juergen.zimmermann)** opened **[DATACMNS-1133](https://jira.spring.io/browse/DATACMNS-1133?redirect=false)** and commented

We finally migrated our microservices to Spring Cloud Finchley.M1, Spring Boot 2.0.M3, Spring Security 5.0.0.M3, and Spring Data Kay.RC2.All Repos are reactive.
However, in our "dev" profile we we are populating a MongoDB database via Jackson2RepositoryPopulatorFactoryBean. Just for this purpose we still have to define (synchroneous) CrudRepositories.
Therefore, we would appreciate if Jackson2RepositoryPopulatorFactoryBean could use our ReactiveCrudRepositories

---

**Affects:** 2.0 RC2 (Kay)

1 votes, 2 watchers

porterwoodward:
After stepping through the code - and watching the `org.springframework.data.repository.support.ReflectionRepositoryInvoker` manage to locate the `.save` method on the reactive mongo repo - but then fail to _subscribe_ to the resulting mono/result it makes sense that the operations are never committed.

```
	protected RepositoryInvoker createInvoker(RepositoryInformation information, Object repository) {

		if (repository instanceof PagingAndSortingRepository) {
			return new PagingAndSortingRepositoryInvoker((PagingAndSortingRepository<Object, Object>) repository, information,
					conversionService);
		} else if (repository instanceof CrudRepository) {
			return new CrudRepositoryInvoker((CrudRepository<Object, Object>) repository, information, conversionService);
		} else {
			return new ReflectionRepositoryInvoker(repository, information, conversionService);
		}
	}
```
Perhaps a check for `instanceof ReactiveSortingRepository` where the return is a new, sub-class of the `ReflectionRepositoryInvoker`, `ReflectionReactiveRepositoryInvoker` - where that invoker calls the terminal method on the Mono<> returned from the save method(s)?

dspitzer:
This took me by surprise as well and cost me quite a few minutes to figure this out. Seemingly for a very long time the populator just doesn't work with reactive repositories.

rolag-it:
Hello,
I'm working on a fix, see my pull request #2543

nerg4l:
Related #2558

[...]

PR #2543, opened 2022-02-04
Add support for Reactive Repositories in Jackson2RepositoryPopulatorFactoryBean

rolag-it:
Related to issue #1575 
This fix adds the interface `RepositoryPopulatorPersistCallback` as callback to `RepositoryPopulator` on entity saving.
So `Jackson2RepositoryPopulatorFactoryBean` can work just adding a bean implementing `RepositoryPopulatorPersistCallback` that calls `block()` on received resulting `Mono` object.
Something like this:

```
                @Bean
		public RepositoryPopulatorPersistCallback persistCallback() {
			return (r,i,o)->{
				if(o instanceof Mono<?>) {
					((Mono<?>) o).block();
				}
			};
		}

		@Bean
		public Jackson2RepositoryPopulatorFactoryBean getRespositoryPopulator() {
			var factory = new Jackson2RepositoryPopulatorFactoryBean();
			factory.setResources(new Resource[] { new ClassPathResource("data.json") });

			return factory;
		}
```


mp911de:
Thanks for your suggestion. After revising the idea we came to the conclusion that we want to avoid multi-threaded arrangements during the application container startup. We instead recommend using imperative repositories for initialization purposes. Therefore I created #2558 to track those efforts.

PR #2566, opened 2022-03-01
Add support for property-specific converters

christophstrobl:
This PR will introduce common infrastructure that allows identifying and serving converters that are property specific. Other than existing converters configured via custom conversions, the property based variant will only be applied to a specific property of a type, instead of targeting all properties of a certain type.

`PropertyValueConverter` implementations will be provided a `ValueConversionContext` that gives access to the target property itself and may be enhanced by the store specific implementation.

```java
class Person {
  // ...
  @ValueConverter(EncryptedValueConverter.class)
  String ssn;
}

class EncryptedValueConverter implements PropertyValueConverter<String, byte[], ValueConversionContext> {

  @Override
  public String read(byte[] value, ValueConversionContext context) {
    // ...
  }

  @Override
  public byte[] write(String value, ValueConversionContext context) {
    // ...
  }
}
```

In addition to the annotation based approach it is possible to programatically register converters for certain properties

```java
PropertyValueConverterRegistrar registrar = new PropertyValueConverterRegistrar();

// untyped registration
registrar.registerConverter(Address.class, "street", new PropertyValueConverter() { ... });

// type safe registration
registrar.registerConverter(Person.class, Person::getSsn())
  .writing(value -> encrypt(value))
  .reading(value -> decrypt(value));
```

### DIFF
--- a/src/main/java/org/springframework/data/convert/PropertyValueConverter.java
+++ b/src/main/java/org/springframework/data/convert/PropertyValueConverter.java
@@ -21,11 +21,11 @@ import org.springframework.data.mapping.PersistentProperty;
 import org.springframework.lang.Nullable;
 
 /**
- * {@link PropertyValueConverter} provides a symmetric way of converting certain properties from domain to store
- * specific values.
+ * {@link PropertyValueConverter} provides a symmetric way of converting certain properties from domain to
+ * store-specific values.
  * <p>
  * A {@link PropertyValueConverter} is, other than a {@link ReadingConverter} or {@link WritingConverter}, only applied
- * to special annotated fields which allows a fine grained conversion of certain values within a specific context.
+ * to special annotated fields which allows a fine-grained conversion of certain values within a specific context.
  *
  * @author Christoph Strobl
  * @param <A> domain specific type.
@@ -36,7 +36,7 @@ import org.springframework.lang.Nullable;
 public interface PropertyValueConverter<A, B, C extends ValueConversionContext<? extends PersistentProperty<?>>> {
 
 	/**
-	 * Convert the given store specific value into it's domain value representation. Typically a {@literal read}
+	 * Convert the given store specific value into it's domain value representation. Typically, a {@literal read}
 	 * operation.
 	 *
 	 * @param value can be {@literal null}.
@@ -47,7 +47,7 @@ public interface PropertyValueConverter<A, B, C extends ValueConversionContext<?
 	A read(@Nullable B value, C context);
 
 	/**
-	 * Convert the given domain specific value into it's native store representation. Typically a {@literal write}
+	 * Convert the given domain specific value into it's native store representation. Typically, a {@literal write}
 	 * operation.
 	 *
 	 * @param value can be {@literal null}.
@@ -67,13 +67,15 @@ public interface PropertyValueConverter<A, B, C extends ValueConversionContext<?
 
 		INSTANCE;
 
+		@Nullable
 		@Override
-		public Object read(Object value, ValueConversionContext context) {
+		public Object read(@Nullable Object value, ValueConversionContext context) {
 			return value;
 		}
 
+		@Nullable
 		@Override
-		public Object write(Object value, ValueConversionContext context) {
+		public Object write(@Nullable Object value, ValueConversionContext context) {
 			return value;
 		}
 	}
@@ -96,13 +98,15 @@ public interface PropertyValueConverter<A, B, C extends ValueConversionContext<?
 			this.reader = reader;
 		}
 
+		@Nullable
 		@Override
-		public B write(A value, ValueConversionContext<P> context) {
+		public B write(@Nullable A value, ValueConversionContext<P> context) {
 			return writer.apply(value, context);
 		}
 
+		@Nullable
 		@Override
-		public A read(B value, ValueConversionContext<P> context) {
+		public A read(@Nullable B value, ValueConversionContext<P> context) {
 			return reader.apply(value, context);
 		}
 	}
--- a/src/main/java/org/springframework/data/convert/PropertyValueConverterRegistrar.java
+++ b/src/main/java/org/springframework/data/convert/PropertyValueConverterRegistrar.java
@@ -29,7 +29,7 @@ import org.springframework.util.Assert;
  * that can be used with {@link PropertyValueConversions}.
  * <p>
  * It is possible to register type safe converters via {@link #registerConverter(Class, Function)}
- * 
+ *
  * <pre class="code">
  * registrar.registerConverter(Person.class, Person::getName) //
  * 		.writing(StringConverter::encrypt) //
@@ -58,7 +58,7 @@ public class PropertyValueConverterRegistrar<P extends PersistentProperty<P>> {
 		String propertyName = MethodInvocationRecorder.forProxyOf(type).record(property).getPropertyPath()
 				.orElseThrow(() -> new IllegalArgumentException("Cannot obtain property name!"));
 
-		return new WritingConverterRegistrationBuilder<T, S, P>(type, propertyName, this);
+		return new WritingConverterRegistrationBuilder<>(type, propertyName, this);
 	}
 
 	/**
@@ -71,8 +71,8 @@ public class PropertyValueConverterRegistrar<P extends PersistentProperty<P>> {
 	 * @return will never be {@literal null}.
 	 */
 	public <T, S> WritingConverterRegistrationBuilder<T, S, P> registerConverter(Class<T> type, String propertyName,
-			Class<S> propertyType) {
-		return new WritingConverterRegistrationBuilder<T, S, P>(type, propertyName, this);
+			@SuppressWarnings("unused") Class<S> propertyType) {
+		return new WritingConverterRegistrationBuilder<>(type, propertyName, this);
 	}
 
 	/**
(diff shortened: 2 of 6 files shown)


src/main/java/org/springframework/data/convert/CustomConversions.java
```diff
@@ -172,6 +185,36 @@ public void registerConvertersIn(ConverterRegistry conversionService) {
...
+	 *
+	 * @param property must not be {@literal null}. param <A> domain specific type
+	 * @param <B> store native type
+	 * @param <C> conversion context type
+	 * @return the suitable {@link PropertyValueConverter} or {@literal null} if none available.
+	 * @see PropertyValueConversions#getValueConverter(PersistentProperty)
+	 * @since 2.7
+	 */
+	@Nullable
+	public <A, B, C extends PersistentProperty<C>, D extends ValueConversionContext<C>> PropertyValueConverter<A, B, D> getPropertyValueConverter(
```
mp911de, line 213: Let's make the `getPropertyValueConverter()` method non-nullable as we have `hasPropertyValueConverter(…)` methods to check for the presence of converters.

src/main/java/org/springframework/data/convert/CustomConversions.java
```diff
@@ -902,9 +946,30 @@ public ConverterConfiguration(StoreConversions storeConversions, List<?> userCon
...
+		 *
+		 * @param storeConversions must not be {@literal null}.
+		 * @param userConverters must not be {@literal null} use {@link Collections#emptyList()} instead.
+		 * @param converterRegistrationFilter must not be {@literal null}.
+		 * @param propertyValueConversions can be {@literal null}.
+		 * @since 2.7
+		 */
+		public ConverterConfiguration(StoreConversions storeConversions, List<?> userConverters,
+				Predicate<ConvertiblePair> converterRegistrationFilter,
+				@Nullable PropertyValueConversions propertyValueConversions) {
```
mp911de, line 967: Having a null-object for `PropertyValueConversions` would allow to avoid nullable constructor args.

src/main/java/org/springframework/data/convert/PropertyValueConversions.java
```diff
@@ -0,0 +1,70 @@
...
+	/**
+	 * Get the {@link PropertyValueConverter} for the given {@literal property} if present.
+	 *
+	 * @param property must not be {@literal null}. param <A> domain specific type
+	 * @param <B> store native type
+	 * @param <C> conversion context type
+	 * @return the suitable {@link PropertyValueConverter} or {@literal null} if none available.
+	 */
+	@Nullable
+	<A, B, C extends PersistentProperty<C>, D extends ValueConversionContext<C>> PropertyValueConverter<A, B, D> getValueConverter(
```
mp911de, line 54: Let's make the `getPropertyValueConverter()` method non-nullable as we have `hasPropertyValueConverter(…)` methods to check for the presence of converters.

src/main/java/org/springframework/data/convert/PropertyValueConversions.java
```diff
@@ -0,0 +1,70 @@
...
+			C property);
+
+	/**
+	 * Helper that allows to create {@link PropertyValueConversions} instance with the configured
+	 * {@link PropertyValueConverter converters} provided via the given callback.
+	 */
+	static <P extends PersistentProperty<P>> PropertyValueConversions simple(
```
mp911de, line 61: How about naming this method `create`? `simple` implies there could be also a `complex` variant.

src/main/java/org/springframework/data/convert/PropertyValueConverter.java
```diff
@@ -0,0 +1,109 @@
...
+	/**
+	 * Convert the given store specific value into it's domain value representation. Typically a {@literal read}
+	 * operation.
+	 *
+	 * @param value can be {@literal null}.
+	 * @param context never {@literal null}.
+	 * @return the converted value. Can be {@literal null}.
+	 */
+	@Nullable
+	A read(@Nullable B value, C context);
```
mp911de, line 47: Didn't we want to have `default` methods or similar to avoid nullability for the value to read/to write?

src/main/java/org/springframework/data/convert/PropertyValueConverterFactories.java
```diff
@@ -0,0 +1,240 @@
...
+			try {
+				return beanFactory.getBean(converterType);
+			} catch (NoSuchBeanDefinitionException exception) {
+
+				if (beanFactory instanceof AutowireCapableBeanFactory) {
+					return (PropertyValueConverter<S, T, C>) ((AutowireCapableBeanFactory) beanFactory).createBean(converterType,
+							AutowireCapableBeanFactory.AUTOWIRE_CONSTRUCTOR, false);
+				}
+			}
+			return null;
```
mp911de, line 138: Shouldn't we throw an exception here if we cannot instantiate the converter? A silent fall-through doesn't seem appropriate.

src/main/java/org/springframework/data/convert/PropertyValueConverterFactories.java
```diff
@@ -0,0 +1,240 @@
...
+
+			Optional<PropertyValueConverter<?, ?, ? extends ValueConversionContext<?>>> converter = cache.get(converterType);
+
+			return converter != null ? (PropertyValueConverter) converter.orElse(null)
+					: cache.cache(converterType, delegate.getConverter(converterType));
+		}
+
+		static class Cache {
+
+			Map<PersistentProperty<?>, Optional<PropertyValueConverter<?, ?, ? extends ValueConversionContext<?>>>> perPropertyCache = new HashMap<>();
```
mp911de, line 215: We either need synchronization or should use `ConcurrentHashMap` to avoid concurrent modification exceptions.

src/main/java/org/springframework/data/convert/PropertyValueConverterFactory.java
```diff
@@ -0,0 +1,149 @@
...
+	 * Get the converter by its type.
+	 *
+	 * @param converterType must not be {@literal null}.
+	 * @param <A> domain specific type.
+	 * @param <B> store native type.
+	 * @param <C> value conversion context to use.
+	 * @return
+	 */
+	@Nullable
+	<A, B, C extends ValueConversionContext<?>> PropertyValueConverter<A, B, C> getConverter(
```
mp911de, line 74: Let's make the `getPropertyValueConverter()` method non-nullable and introduce a `hasConverter` method to check for the presence of converters.

src/main/java/org/springframework/data/convert/PropertyValueConverterRegistrar.java
```diff
@@ -0,0 +1,191 @@
...
+			target.registerConverter(entry.getKey().type, entry.getKey().path, entry.getValue());
+		});
+	}
+
+	/**
+	 * Helper to build up a fluent registration API starting on
+	 *
+	 * @author Oliver Drotbohm
+	 */
+	static class WritingConverterRegistrationBuilder<T, S, P extends PersistentProperty<P>> {
```
mp911de, line 123: We need to increase visibility so that builders can be used from outside of the package.

src/main/java/org/springframework/data/convert/ValueConverter.java
```diff
@@ -0,0 +1,53 @@
...
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ValueConverter {
+
+	/**
+	 * The {@link PropertyValueConverter} type handling the value conversion of the annotated property.
+	 *
+	 * @return the configured {@link PropertyValueConverter}. {@link ObjectToObjectPropertyValueConverter} by default.
+	 */
+	Class<? extends PropertyValueConverter> value() default ObjectToObjectPropertyValueConverter.class;
```
mp911de, line 51: Can we make the `value` mandatory?

src/main/java/org/springframework/data/convert/ValueConverterRegistry.java
```diff
@@ -0,0 +1,77 @@
...
+	 * Obtain the converter registered for the given type, path combination or {@literal null} if none defined.
+	 *
+	 * @param type the target type. Must not be {@literal null}.
+	 * @param path the property name. Must not be {@literal null}.
+	 * @param <S>
+	 * @param <T>
+	 * @return {@literal null} if no converter present for the given type/path combination.
+	 */
+	@Nullable
+	<S, T> PropertyValueConverter<S, T, ? extends ValueConversionContext<P>> getConverter(Class<?> type, String path);
```
mp911de, line 50: Let's make the `getConverter()` method non-nullable as we have `containsConverterFor(…)` methods to check for the presence of converters.

src/main/java/org/springframework/data/convert/ValueConverterRegistry.java
```diff
@@ -0,0 +1,77 @@
...
+	 */
+	boolean isEmpty();
+
+	/**
+	 * Obtain a simple {@link ValueConverterRegistry}.
+	 *
+	 * @param <T>
+	 * @return new instance of {@link ValueConverterRegistry}.
+	 */
+	static <T extends PersistentProperty<T>> ValueConverterRegistry<T> simple() {
```
mp911de, line 74: Let's call this `create` as we have no other variants.

src/main/java/org/springframework/data/convert/PropertyValueConverterRegistrar.java
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	/**
+	 * A helper to build a fluent API to register how to read a database value into a domain object property.
+	 *
+	 * @author Oliver Drotbohm
+	 */
+	static class ReadingConverterRegistrationBuilder<T, S, R, P extends PersistentProperty<P>> {
+
+		private WritingConverterRegistrationBuilder<T, S, P> origin;
```
mp911de, line 161: Generics seem off here as `T` refers to the underlying type. Later on, we call `registration.accept` where generics do not match.

mp911de:
That's merged, polished, and forward-ported now.

PR #2570, opened 2022-03-02
Consider annotated methods in `AnnotationRevisionMetadata`

### LINKED ISSUE #2569: `AnnotationRevisionMetadata` should consider annotations on methods
Based on spring-projects/spring-data-envers#313 we should try to detect annotations on methods as well as e.g. `org.hibernate.envers.RevisionNumber` can be used to annotate methods.

mp911de:
We now detect annotated methods.

_The polishing commit should not be backported._

Closes #2569

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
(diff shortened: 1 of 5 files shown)


schauder:
Merged, and ported back and forward.

PR #2571, opened 2022-03-07
Allow using Range with `Comparator`

mp911de:
Remove `Comparable` requirement for general `Range` usage. `Comparable` is only required for the `contains(…)` check. Alternatively, accept a `Comparator`.

### DIFF
--- a/src/main/java/org/springframework/data/domain/Range.java
+++ b/src/main/java/org/springframework/data/domain/Range.java
@@ -185,21 +185,14 @@ public final class Range<T> {
 	 * @return
 	 */
 	@SuppressWarnings({ "unchecked" })
-	public <V extends Comparable<V>> boolean contains(V value) {
+	public boolean contains(Comparable<T> value) {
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/domain/Range.java
```diff
@@ -183,16 +184,41 @@ private Range(Bound<T> lowerBound, Bound<T> upperBound) {
 	 * @param value must not be {@literal null}.
 	 * @return
 	 */
-	public boolean contains(T value) {
+	@SuppressWarnings({ "unchecked" })
+	public <V extends Comparable<V>> boolean contains(V value) {
```
odrotbohm, line 188: Shouldn't this just delegate to `….contains(…, Comparator)` with a comparator using the `Comparable` methods? Otherwise, the code implementing the actual logic is duplicated between the methods?

What's also a bit of a pity is that you can now call `..contains("someString")` on a `Range<Integer>`. Unfortunately, there doesn't seem to be a way to tighten the generics declaration in that regard.

PR #2591, opened 2022-04-04
Eagerly initialize `SimplePropertyValueConversions`

christophstrobl:
Resolves: #2590


mp911de:

PR #2596, opened 2022-04-06
Introduce `PropertiesBasedNamedQueriesFactoryBean` to reduce bean indirections

### LINKED ISSUE #2584: Refine inner beans arrangement
When defining repository factory beans, we make use of inner beans to e.g. load named queries. Ideally, we register a single bean that gets reused across multiple factory beans.

mp911de:
We now provide `PropertiesBasedNamedQueriesFactoryBean` to create `PropertiesBasedNamedQueries` directly bypassing indirections through `PropertiesFactoryBean` and `PropertiesBasedNamedQueries`.

Closes #2584

### DIFF
--- a/src/main/java/org/springframework/data/repository/config/RepositoryBeanDefinitionBuilder.java
+++ b/src/main/java/org/springframework/data/repository/config/RepositoryBeanDefinitionBuilder.java
@@ -16,6 +16,7 @@
 package org.springframework.data.repository.config;
 
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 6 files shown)


src/main/java/org/springframework/data/repository/config/RepositoryBeanDefinitionBuilder.java
```diff
@@ -128,8 +131,9 @@ public BeanDefinitionBuilder build(RepositoryConfiguration<?> configuration) {
 
 		fragmentsBuilder.addConstructorArgValue(fragmentBeanNames);
 
-		builder.addPropertyValue("repositoryFragments",
-				ParsingUtils.getSourceBeanDefinition(fragmentsBuilder, configuration.getSource()));
+		String fragmentsBeanName = BeanDefinitionReaderUtils.uniqueBeanName(extension.getModuleName().toLowerCase(Locale.ROOT) + ".repo-fragments", registry);
```
mp911de, line 134: It would make sense to include the repository name for easier correlation as the fragments are correlated to an individual repository.

Issue #2064, opened 2019-12-15
Provide a dedicated read-only repository interface [DATACMNS-1638]


spring-projects-issues:
**[Vitor Carvalho](https://jira.spring.io/secure/ViewProfile.jspa?name=vitorfec)** opened **[DATACMNS-1638](https://jira.spring.io/browse/DATACMNS-1638?redirect=false)** and commented

Whenever a developer needs to create a read only repository, it has to extend the Repository interface and then add all the read methods from CrudRepository to it.

It would be useful to have a ReadRepository similar to CrudRepository with all the read only methods.

```java
Optional<T> findById(ID id);

boolean existsById(ID id);

Iterable<T> findAll();

Iterable<T> findAllById(Iterable<ID> ids);

long count();
```

---
No further details from [DATACMNS-1638](https://jira.spring.io/browse/DATACMNS-1638?redirect=false)

gregturn:
Duplicates #2610.

gregturn:
Based on discussions with the team, we have decided NOT to implement this. However, you are free to implement it locally in your own project. 

Check out this [part of the ref docs](https://docs.spring.io/spring-data/commons/docs/3.0.0-M3/reference/html/#repositories.definition), where you can see how to define a base repository of your own, e.g. ReadOnlyRepository (or ReadingRepository) where you could have only read-based operations.

PR #2611, opened 2022-04-20
Introduce ReadOnlyRepository between Repository and CrudRepository.

### LINKED ISSUE #2610: Introduce ReadOnlyRepository between Repository and CrudRepository
**[Ted Bergeron](https://jira.spring.io/secure/ViewProfile.jspa?name=tedberg)** opened **[DATAJPA-541](https://jira.spring.io/browse/DATAJPA-541?redirect=false)** and commented

ReadOnlyRepository can be created as an extension of Repository for an app, but this seems like basic functionality that should be included.  A sticking point mentioned in [DATAREST-299](https://jira.spring.io/browse/DATAREST-299) is that DomainClassConverter only works if your entity repository extends CrudRepository.

I am using DomainClassConverter against entities marked `@Immutable`, which are just reference data objects.  If ReadOnlyRepository were refactored out into a parent interface, it seems that DomainClassConverter should work against it instead of CrudRepository

---

**Affects:** 1.5.2 (Codd SR2)

**Issue Links:**
- [DATACMNS-102](https://jira.spring.io/browse/DATACMNS-102) Allow Repositories to be composed of an arbitrary number of implementation classes

gregturn:
Given it wasn't hard, I went ahead and coded it and made sure existing tests worked properly.  If you agree to it, we can merge it. If we don't want it, then we can drop it.

Closes #2610, #2064

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>3.0.0-SNAPSHOT</version>
+	<version>3.0.0-gh-2610-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 
--- a/src/main/java/org/springframework/data/repository/CrudRepository.java
+++ b/src/main/java/org/springframework/data/repository/CrudRepository.java
@@ -25,7 +25,7 @@ import java.util.Optional;
  * @author Jens Schauder
  */
 @NoRepositoryBean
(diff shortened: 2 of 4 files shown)


mp911de:
With the current arrangement, we would also need `List`-based variants of the repositories. Let's discuss the approach with the team as we never were convinced that another repository interface and the conceptual complexity associated with such a change would warrant the benefit. One could subclass `ReadOnlyRepository` and add a `@Modifying` method and easily break the concept.

gregturn:
Based on discussions, I am withdrawing this PR.

PR #2615, opened 2022-04-27
Expose BeanFactory to RepositoryFactorySupport implementations. 

gregturn:
This is needed to close https://github.com/spring-projects/spring-data-jpa/pull/2513.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>3.0.0-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


mp911de:
Let's keep that change local to the JPA module. Since `RepositoryFactorySupport` implements `BeanFactoryAware` already, it would make sense to override the `setBeanFactory` method in the JPA Repository Factory.

gregturn:
I don’t follow. This is adding getBeanFactory, not setBeanFactory.

mp911de:
What I mean is, that the method that is used to set the bean factory can be overridden in the JPA repository factory to capture the actual value (`BeanFactory` instance).

gregturn:
Withdrawn.

PR #2616, opened 2022-04-28
Support reactive repositories in `ResourceReaderRepositoryPopulator` 

### LINKED ISSUE #2558: Ensure that `AbstractRepositoryPopulatorFactoryBean` uses imperative repositories only
To ensure we do not introduce threading during the application container startup we want to use imperative repositories only.

mp911de:
We now support reactive repositories in addition to imperative repositories when populating repositories from a resource reader.

Instead of `RepositoryInvokerFactory`, we now use a `RepositoryPersisterFactory` to avoid introducing reactive support to `RepositoryInvokerFactory`.

Closes #2558

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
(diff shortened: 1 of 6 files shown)


odrotbohm:
That's polished and merged.

PR #2624, opened 2022-05-05
Introduce common AOT infrastructure.

christophstrobl:
The commons module hosts the basic infrastructure for contributing data related bits to the AOT processing.

We introduced `ManagedTypes` as an abstraction that allows to define what was up till now defined via the `initialEntitySet` in the `MappingContext`. 
`ManagedTypes` may be provided per store module as a regular bean in the application context, if not present a lazy evaluating variant will be registered, picking up types annotated with one of the store specific `identifyingAnnotations` within the repository base packages.
This allows store implementations to provide an `AotManagedTypesPostProcessor` picking up the bean to render additional configuration (reflection, proxy,...) for the domain types present.

`RepositoryConfigurationExtension.getAotPostProcessor()` defines the type of the `AotContributingRepositoryBeanPostProcessor` to be used by a store module. By default the variant provided by data commons will contribute proxy/reflection configuration for all types reachable from the repository interface. This includes the domain & id types as well as those returned by repository finder methods identifying potential projection proxies.


src/main/java/org/springframework/data/aot/TypeUtils.java
```diff
@@ -0,0 +1,253 @@
...
+import org.springframework.core.annotation.MergedAnnotation;
+import org.springframework.core.annotation.MergedAnnotations;
+import org.springframework.core.annotation.MergedAnnotations.SearchStrategy;
+import org.springframework.core.annotation.RepeatableContainers;
+import org.springframework.util.ObjectUtils;
+
+/**
+ * @author Christoph Strobl
+ */
+public class TypeUtils {
```
jxblum, line 41: Perhaps this class can be moved to org.springframework.data.util.

I don't see anything in this class that is specific to the native AOT processing infrastructure classes in Spring Data (i.e. org.springframework.data.aot).

src/main/java/org/springframework/data/ManagedTypes.java
```diff
@@ -0,0 +1,64 @@
...
+
+	static ManagedTypes of(Iterable<? extends Class<?>> types) {
+		return types::forEach;
+	}
+
+	static ManagedTypes of(Stream<? extends Class<?>> types) {
+		return types::forEach;
+	}
+
+	static ManagedTypes o(Supplier<Iterable<? extends Class<?>>> dataProvider) {
```
jxblum, line 52: Misspelled. No matter, I fixed in a polish commit.

src/main/java/org/springframework/data/aot/TypeCollector.java
```diff
@@ -0,0 +1,243 @@
...
+			return false;
+		}
+		if (field.getDeclaringClass().getPackageName().startsWith("java.")) {
+			return false;
+		}
+		return true;
+	};
+
+	public TypeCollector filterFields(Predicate<Field> filter) {
+		this.fieldFilter = filter.and(filter);
```
jxblum, line 86: Just curious if you meant to prepend here?  I ask because the filterTypes(:Predicate) method below "appends".

src/main/java/org/springframework/data/aot/TypeCollector.java
```diff
@@ -0,0 +1,243 @@
...
+		private List<Class<?>> collect() {
+			List<Class<?>> target = new ArrayList<>();
+			forEach(it -> target.add(it.toClass()));
+			return target;
+		}
+	}
+
+	static class InspectionCache {
+
+		private final Map<String, ResolvableType> mutableCache = new LinkedHashMap<>();
```
jxblum, line 221: Maybe a WeakHashMap?

Issue #2634, opened 2022-05-17
Add infrastructure to provide/identify managed types.


christophstrobl:
The `MappingContext` allows to define a set of initial entities used to compute metadata early instead of first encounter. Those types are identified in multiple ways, depending if set by the user, boot or data itself. The latter two make use of the identifying annotations provided by the store module implementation.
Being buried within the mapping context, potential AOT processing is hardly possible. 

**Desired State:**
- Each store registers a bean holding references to the type names it needs to know about if not provided externally
- Mapping context obtains the bean and uses the given types to initialize the mapping context.
- Public infrastructure that does the entity scanning so we do not have different implementations in data and boot.

PR #2635, opened 2022-05-18
Introduce ManagedTypes.

christophstrobl:
See: #2634


src/main/java/org/springframework/data/repository/config/RepositoryConfigurationExtension.java
```diff
@@ -65,6 +79,17 @@ <T extends RepositoryConfigurationSource> Collection<RepositoryConfiguration<T>>
...
 	String getRepositoryFactoryBeanClassName();
 
+	/**
+	 * Return the annotations to scan domain types for when evaluating repository interfaces for store assignment. Modules
+	 * should return the annotations that identify a domain type as managed by the store explicitly.
+	 *
+	 * @return
+	 * @since 3.0
+	 */
+	default Collection<Class<? extends Annotation>> getIdentifyingAnnotations() {
```
mp911de, line 89: It feels a bit weird handing out identifying annotations to the outside of the config extension as the managed type scanning vs. repository scanning are located at two different abstractions.

Repository type scanning happens inside the `RepositoryConfigurationExtension`. We still need to merge repository-managed types with the ones that we discover through the classpath scan.

How about returning a `RepositoryConfigurations` object from `getRepositoryConfigurations` that wraps `Collection<RepositoryConfiguration<T>>` and provides a `ManagedTypes` object?

Doing so would allow keeping all scanning-related operations within the extension.

src/main/java/org/springframework/data/util/TypeScanner.java
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	/**
+	 * Collects the {@link String names} of packages to scan.
+	 *
+	 * @param packageNames array of {@link String package names} to scan; Must not be {@literal null}.
+	 * @return new instance of {@link TypeScanner}.
+	 * @see #scanPackages(Collection)
+	 */
+	default TypeScanner scanPackages(String... packageNames) {
```
mp911de, line 81: `scanPackages` should be renamed to a name that is less imperative and expresses rather that this is a config option `packagesToScan` or something like that, so that the imperative names (`forEach`, `collectAsSet`) remain as terminal methods.

jxblum: Taking this 1 step further, perhaps all the `typeScanner(..`) factory methods could be renamed to `from(..)` and the `scanPackages(..)` method could be renamed to `inPackages(..)`.

src/main/java/org/springframework/data/repository/config/RepositoryConfigurationDelegate.java
```diff
@@ -203,9 +209,49 @@ public List<BeanComponentDefinition> registerRepositoriesIn(BeanDefinitionRegist
...
+		registerManagedTypes(registry, extension,
+				() -> configurations.stream().flatMap(it -> it.getBasePackages().stream()).collect(Collectors.toSet()));
 		return definitions;
 	}
 
+	protected void registerManagedTypes(BeanDefinitionRegistry registry, RepositoryConfigurationExtension extension,
+			Supplier<Set<String>> basePackages) {
+
+		String beanName = String.format("%s.managed-types", extension.getModulePrefix());
+		if (!registry.isBeanNameInUse(beanName)) {
```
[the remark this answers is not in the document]
christophstrobl: resolved by not not scanning for types here. (see comment below).

src/main/java/org/springframework/data/util/TypeScanner.java
christophstrobl: considered - `TypeScanner` is likely to be dropped from this PR and move to another one.

mp911de:

Issue #2481, opened 2021-10-12
Notify `*Aware` interfaces using custom type filters through `Enable…Repositories#includeFilters`


f-cramer:
The javadoc for `ComponentScan.Filter#classes` specifies that custom instances of `TypeFilter` can implement any of 4 `*Aware` interfaces, whose methods should be called before matching.
When specifying a filter in `EnableJpaRepositories#includeFilters`, the methods of those interfaces are _not_ called. Construtor autowiring cannot be used either, because filter instances are created using `BeanUtils#instantiateClass(Class<?>)`.

mp911de:
I moved this ticket into Spring Data Commons as the repository discovery is part of our common infrastructure.

mp911de:
I filed https://github.com/spring-projects/spring-framework/issues/27553 asking for reuse of Framework components for filter annotation parsing.

PR #2647, opened 2022-06-21
Use TypeFilterUtils to create TypeFilters

f-cramer:
Closes gh-2481


mp911de:

PR #2652, opened 2022-07-12
Specify title to `auditing.auditor-aware`.

michael-simons:
If `auditing.adoc` is included in the rendered documentation, the sentences does not make too much sense (see screenshot).

Ohne Titel


mp911de:

Issue #2662, opened 2022-07-21
Make support of XML namespaces optional in the documentation


schauder:
This is required for https://github.com/spring-projects/spring-data-relational/issues/1290

PR #2664, opened 2022-07-22
Make reference to XML namespace optional.

schauder:
Not all modules have a XML namespace.
Therefore it is now possible to exclude parts that reference the XML namespace from the documentation

By default the parts referring to the XML namespace are included just as before.
But they will get skipped if the `index.adoc` of the module defines the attribute `include-xml-namespaces` as `false` by including

```
:include-xml-namespaces: false
```

Closes https://github.com/spring-projects/spring-data-commons/issues/2662
See https://github.com/spring-projects/spring-data-relational/issues/1290

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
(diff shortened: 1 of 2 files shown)


schauder:
Merged.

PR #2673, opened 2022-08-03
Clarify behaviour of when save or delete operations result in a NOOP.

### LINKED ISSUE #2651: missing description of thrown EmptyResultDataAccessException in deleteById
The implementation of CrudRepository.deleteById throws an EmptyResultDataAccessException when no entity for the id exists. This is not described in the javadoc.

See interface https://github.com/spring-projects/spring-data-commons/blob/main/src/main/java/org/springframework/data/repository/CrudRepository.java#L97-L103
```java
	/**
	 * Deletes the entity with the given id.
	 *
	 * @param id must not be {@literal null}.
	 * @throws IllegalArgumentException in case the given {@literal id} is {@literal null}
	 */
	void deleteById(ID id);
```

See implementation of the interface https://github.com/spring-projects/spring-data-jpa/blob/main/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/support/SimpleJpaRepository.java#L164-L172
```java
	@Transactional
	@Override
	public void deleteById(ID id) {

		Assert.notNull(id, ID_MUST_NOT_BE_NULL);

		delete(findById(id).orElseThrow(() -> new EmptyResultDataAccessException(
				String.format("No %s entity with id %s exists", entityInformation.getJavaType(), id), 1)));
	}
```

The `EmptyResultDataAccessException` is not part of spring-data-commons. So maybe that's the reason it's not mentioned here.

But can we add a hint to this behavior somehow? From the current javadoc I would expect no exception if no entity was found, similar to a sql deletion statement.

schauder:
Behaviour of implementing modules need to adapt to the new clarification.

Closes #2651

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>3.0.0-SNAPSHOT</version>
+	<version>3.0.0-2651-failed-deletes-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 	<description>Core Spring concepts underpinning every Spring Data module.</description>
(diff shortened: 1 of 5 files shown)


mp911de:

Issue #2677, opened 2022-08-18
Provide invocation arguments for RepositoryMethodInvocation


roookeee:
`RepositoryTagsProvider` of `spring-boot-actuator` is based on `RepositoryMethodInvocation` which is defined here.
We would like to tag our repository metrics based on the passed arguments to provide more granular metrics. For example: We have a query that takes a `List<String>` and we want to track metrics for each size of the given parameter, as the query has / may have difference performance characteristics when being passed 1 vs. 100 ids. This would skew the gathered metrics and overall make the metrics unusable as they can't be interpreted.

From reading the code `RepositoryMethodInvocation` could be passed the arguments of the actual invocation which would then be available to the `RepositoryTagsProvider`. Is there any downside to this? Are the arguments only the repository method call arguments or are there some framework related arguments present?

Thank you for your time. I would be willing to do a PR for this if you agree that this should be done.

mp911de:
Generally speaking, providing arguments would be possible but any argument-post-processing (unwrapping `Optional`, `Future`) would be up to your code. Also, any reactive arguments (`Mono<Person>`) would be passed on as-is and you are not able to consume those parameters in an non-blocking way.

Feel free to submit a pull request.

roookeee:
I created a PR for this issue [here](https://github.com/spring-projects/spring-data-commons/pull/2681/). Let me know if you want any adjustments via comments in there

PR #2681, opened 2022-08-26
Provide invocation arguments for RepositoryMethodInvocation

roookeee:
See #2677 for more information about this PR.

### DIFF
--- a/src/main/java/org/springframework/data/repository/core/support/RepositoryInvocationMulticaster.java
+++ b/src/main/java/org/springframework/data/repository/core/support/RepositoryInvocationMulticaster.java
@@ -56,7 +56,7 @@ interface RepositoryInvocationMulticaster {
 
 	/**
 	 * {@link RepositoryInvocationMulticaster} implementation that notifies {@link RepositoryMethodInvocationListener}
-	 * upon {@link #notifyListeners(Method, Object[], RepositoryMethodInvocationResult)}.
+	 * upon {@link #notifyListeners(Method, Object[], RepositoryMethodInvocation)}.
 	 *
 	 * @author Mark Paluch
 	 */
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/repository/core/support/RepositoryInvocationMulticaster.java
```diff
@@ -56,7 +56,7 @@ public void notifyListeners(Method method, Object[] args, RepositoryMethodInvoca
 
 	/**
 	 * {@link RepositoryInvocationMulticaster} implementation that notifies {@link RepositoryMethodInvocationListener}
-	 * upon {@link #notifyListeners(Method, Object[], RepositoryMethodInvocationResult)}.
```
roookeee, line 59: I just fixed this broken JavaDoc which I noticed along the way, it's not directly related to this pr.

roookeee:
I put the new `arguments` parameter where it felt "right", let me know if you want it some other way (e.g. always at the end).
`RepositoryInvocationMulticaster` could be refactored to not be passed the `method` and `args` as they are provided by the `RepositoryMethodInvocation` - I didn't want to touch this without any further insight from a maintainer.

src/test/java/org/springframework/data/repository/core/support/RepositoryMethodInvokerUnitTests.java
```diff
@@ -197,6 +197,7 @@ void capturesImperativeSuccessCorrectly() throws Exception {
 
 		assertThat(multicaster.first().getResult().getState()).isEqualTo(State.SUCCESS);
 		assertThat(multicaster.first().getResult().getError()).isNull();
+		assertThat(multicaster.first().getArguments()).isNotNull();
```
roookeee, line 200: I felt like its appropriate to check all state case results should have arguments present, any thoughts on this? I could make this a more explicit check if you want.

PR #2694, opened 2022-10-04
Fall back to canonical constructor in constructor resolution when using Java Records

### LINKED ISSUE #2625: Fall back to canonical constructor in constructor resolution when using Java Records
Hello,

I have an issue when attempting to persist a record.

The following is fine:

```java
@Table("cat")
public record Cat(
        @Id @Column("id") Long id,
        @Column("name") String name) {

    // Instead of using a constructor uses a factory method.
    public static Cat of(String name) {
        // The 'id' is auto-generated by the DB and populated on insert.
        return new Cat(null, name);
    }

}
```

But when I use another constructor instead of a factory method I get an exception. The code:

```java
@Table("dog")

public record Dog(
        @Id @Column("id") Long id,
        @Column("name") String name) {

    /**
     * Convenience constructor which sets the id to null with the expectation that it will
     * be populated when the entity is inserted into the DB (since it is an auto-incremented column).
     */
    public Dog(String name) {
        this(null, name);
    }
}
```

Test which fails that I would except to pass:

```java
@SpringBootTest
public class DogTest {
    private static final Logger log = LoggerFactory.getLogger(DogTest.class);

    @Autowired
    private JdbcAggregateTemplate jdbcAggregateTemplate;

    // THIS TEST FAILS CURRENTLY.
    @Test
    void animal_insertingIntoDb_succeeds() {
        /* Given: A dog. */
        var fido = new Dog("Fido");

        /* When: We try to insert it in to the DB */
        var insertedFido = jdbcAggregateTemplate.insert(fido);

        /* Then: It should populate the ID field */
        assertThat(insertedFido.id()).isNotNull();
    }
}
```

And the exception:

```
Cannot set property id because no setter, wither or copy constructor exists for class com.example.demo.persistence.Dog!
java.lang.IllegalStateException: Cannot set property id because no setter, wither or copy constructor exists for class com.example.demo.persistence.Dog!
	at org.springframework.data.mapping.model.InstantiationAwarePropertyAccessor.setProperty(InstantiationAwarePropertyAccessor.java:113)
	at org.springframework.data.mapping.model.ConvertingPropertyAccessor.setProperty(ConvertingPropertyAccessor.java:64)
	at org.springframework.data.jdbc.core.JdbcAggregateChangeExecutionContext.setIdAndCascadingProperties(JdbcAggregateChangeExecutionContext.java:337)
	at org.springframework.data.jdbc.core.JdbcAggregateChangeExecutionContext.populateIdsIfNecessary(JdbcAggregateChangeExecutionContext.java:305)
	at org.springframework.data.jdbc.core.AggregateChangeExecutor.execute(AggregateChangeExecutor.java:52)
	at org.springframework.data.jdbc.core.JdbcAggregateTemplate.store(JdbcAggregateTemplate.java:339)
	at org.springframework.data.jdbc.core.JdbcAggregateTemplate.insert(JdbcAggregateTemplate.java:167)
	at com.example.demo.persistence.DogTest.animal_insertingIntoDb_succeeds(DogTest.java:26)
```

Attached is a repo you can run to reproduce the problem: [spring-data-jdbc-records-bug.zip](https://github.com/spring-projects/spring-data-commons/files/8639991/spring-data-jdbc-records-bug.zip)

Am I doing something wrong here (missing annotation/etc.) I would expect the framework to be able to pick up the correct constructor for the record but it currently does not.

mp911de:
We now fall back to the canonical constructor of records when we cannot disambiguate which constructor to use.

Also, reflect the Kotlin persistence creator mechanism.

Closes #2625

### DIFF
--- a/src/main/java/org/springframework/data/mapping/model/PreferredConstructorDiscoverer.java
+++ b/src/main/java/org/springframework/data/mapping/model/PreferredConstructorDiscoverer.java
@@ -22,6 +22,7 @@ import kotlin.reflect.jvm.ReflectJvmMapping;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.RecordComponent;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -34,11 +35,11 @@ import org.springframework.data.mapping.Parameter;
 import org.springframework.data.mapping.PersistentEntity;
 import org.springframework.data.mapping.PersistentProperty;
 import org.springframework.data.mapping.PreferredConstructor;
-import org.springframework.data.util.ClassTypeInformation;
 import org.springframework.data.util.KotlinReflectionUtils;
 import org.springframework.data.util.TypeInformation;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 
 /**
  * Helper class to find a {@link PreferredConstructor}.
@@ -49,7 +50,7 @@ import org.springframework.util.Assert;
  * @author Mark Paluch
  * @author Xeno Amess
  */
-public interface PreferredConstructorDiscoverer<T, P extends PersistentProperty<P>> {
+public interface PreferredConstructorDiscoverer {
 
 	/**
 	 * Discovers the {@link PreferredConstructor} for the given type.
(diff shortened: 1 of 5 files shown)


odrotbohm (approved):
LGTM

src/main/java/org/springframework/data/mapping/model/PreferredConstructorDiscoverer.java
```diff
@@ -124,12 +125,45 @@ <T, P extends PersistentProperty<P>> PreferredConstructor<T, P> discover(TypeInf
 					}
 				}
 
+				if (rawOwningType.isRecord() && (candidates.size() > 1 || (noArg != null && !candidates.isEmpty()))) {
```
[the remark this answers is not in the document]
schauder: Thinking even more about it, wouldn't it make more sense to handle the cases in the following order (with noargs added to the candidates):

- no candidate
- single candidate
- records

Also, the whole logic should really fail when there is more than one `PreferredCreator` annotation, shouldn't it? It currently it seems to just take the first.

src/test/java/org/springframework/data/mapping/PreferredConstructorDiscovererUnitTests.java
```diff
@@ -149,6 +150,34 @@ void detectsMetaAnnotatedValueAnnotation() {
...
+			assertThat(ctor.getParameters()).hasSize(2);
+			assertThat(ctor.getParameters().get(0).getRawType()).isEqualTo(Long.class);
+			assertThat(ctor.getParameters().get(1).getRawType()).isEqualTo(String.class);
+		});
+	}
+
+	@Test // GH-2332
+	void detectsAnnotatedRecordConstructor() {
+
+		assertThat(PreferredConstructorDiscoverer.discover(RecordWithPersistenceCreator.class)).satisfies(ctor -> {
```
schauder, line 174: Can we make these either soft asserts, or separate tests?

src/test/java/org/springframework/data/mapping/PreferredConstructorDiscovererUnitTests.java
```diff
@@ -149,6 +150,34 @@ void detectsMetaAnnotatedValueAnnotation() {
 		});
 	}
 
+	@Test // GH-2332
```
schauder, line 153: I'd like to see a test for the behavior of a private all args constructor for Records, because right now I can't easily tell how the code behaves.

schauder (changes requested):

src/test/java/org/springframework/data/mapping/PreferredConstructorDiscovererUnitTests.java
mp911de: Constructors that match the canonical constructor cannot be private. Also, a constructor matching the canonical makes Java backoff from its default constructor creation (same as with public no-arg constructors for regular classes).

src/test/java/org/springframework/data/mapping/PreferredConstructorDiscovererUnitTests.java
mp911de: For whatever reason, the entire test class is written in that way. While I find that style strange (likely that comes from our `Optional` experiments), we could apply some polishing post-RC1.

src/main/java/org/springframework/data/mapping/model/PreferredConstructorDiscoverer.java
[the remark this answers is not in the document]
mp911de: Records resolve the no-candidate case by falling back to the constructor that is always there, we do the same for Kotlin data classes already so it makes sense to align. Selecting the no-args constructor for records is pointless with the idea to mimic `with…` style with our `InstantiationAwarePropertyAccessor`.

PR #2706, opened 2022-10-10
Migrate DependencyTests to ArchUnit

schauder:
Replaces the Degraph based implementation with one based on ArchUnit and reenables the test. 

The test currently fails since the aot code causes cycles.


schauder:
Will add Javadoc and license headers.

schauder:
Done.

mp911de:

Issue #2266, opened 2021-01-03
Make Querydsl classes public to allow custom implementation


Sam-Kruglov:
Hello, I have a need for a project to allow something more complex than Web QueryDSL, i.e. I need more operations than just "equals", like: ne, gt, lt, having OR clause etc.

The query language is actually documented [over here on baeldung](https://www.baeldung.com/rest-api-query-search-or-operation).

So, I decided to go with a single String parameter instead of the multi parameters like the default spring approach, i.e. `?query=country eq US and (city ne bla or state ne xx)`.

The problem I'm facing is that I started with defining my own `QuerydslPredicateArgumentResolver`, then my own `QuerydslPredicateBuilder` but I can't because I need `QuerydslBindings#getPropertyPath` as well as access to `PathInformation` to properly convert stuff into the required type. I don't really have to use `QuerydslBindings` but I'd like to take advantage of its features like excluding properties or creating aliases. For now, I decided to drop query DSL and go with Specifications from spring data.

Would you please consider opening up the things you think could be useful for my usecase? Or, perhaps, providing extension points if I want to have more search operations? 

Please, have a look at [this 3rd party project](https://gt-tech.bitbucket.io/spring-data-querydsl-value-operators/README.html) that does what I want by extending the default Spring approach but I decided not to go with it since it seems there's a lack of support. But you could have an idea from their approach as it doesn't conflict with the existing code unlike mine.

mp911de:
Do you want to submit a pull request?

Sam-Kruglov:
I am no longer involved in this project, so, I don't have a need for it anymore, and, while I think this feature would be really nice, I don't have a lot of time on my hands currently.

I think as of right now, Web QueryDSL is too simple for production usage. But if it would be easy to at least plug-in custom operations, or if there would be some essentials out of the box, many people would consider it. It could replace a lot of code for various custom endpoints any application needs. Maybe it could even be somehow used with GraphQL.

Anyway, I understand if you don't have the capacity to deal with this right now.

mp911de:
We're somewhat reluctant to open up the API therefore we were asking for a pull request to see the scope of the changes. Given that the change is no longer necessary, closing this ticket.

rodrigorodrigues:
Hi @mp911de,

I've created a pull request for this ticket to make easier the customisation of queryDsl.

https://github.com/spring-projects/spring-data-commons/pull/2730

PR #2730, opened 2022-11-09
Add Custom QuerydslPredicateBuilder

rodrigorodrigues:
To facilitate the use of a custom queryDsl - related to https://github.com/spring-projects/spring-data-commons/issues/2266


mp911de:
These are quite some changes. The scope is too extensive. It is easier to follow up with a subclass of `QuerydslPredicateBuilder` on your side and apply the binding customizations there instead of trying to squeeze in such changes into something that wasn't designed that way.

PR #2739, opened 2022-11-30
fix a typo in javadoc

ByungjunYou:
This is a fix to correct a typo in the javadoc of the PersistentEntityIsNewStrategy class.

PersistentEntityIsNewStrategy uses the identifier of the given entity as a fall back property to decide whether the entity is new or not.


mp911de:

Issue #2721, opened 2022-10-19
Consider reflection hints for generated Querydsl `Q` classes


mp911de:
When using Querydsl, we attempt to access generated `Q` classes reflectively. Without including runtime hints, reflective access fails.

It would make sense to consider how and where Q classes could be included into the final artifact.

Related to #2717

PR #2743, opened 2022-12-05
Register reflection hints for Querydsl Q types.

christophstrobl:
This PR introduced support for registering GraalVM native reflection type hints for Querydsl `Q` types that target domain types used in the repository interface declaration.

At this point we only do a simple lookup for `Q` types based in the very same package that implement `EntityPath`.
More advanced configuration (eg. base package and type prefix), potentially available via `EntityPathResolver` are not taken into account as this would require eager bean resolution from the application context, which is likely to trigger additional infrastructure. In this case the user is required to register `Q` types manually.

Closes: #2721

### DIFF
--- a/src/main/java/org/springframework/data/repository/config/RepositoryRegistrationAotContribution.java
+++ b/src/main/java/org/springframework/data/repository/config/RepositoryRegistrationAotContribution.java
@@ -42,14 +42,15 @@ import org.springframework.core.DecoratingProxy;
 import org.springframework.core.ResolvableType;
 import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.data.aot.AotContext;
-import org.springframework.data.util.TypeContributor;
-import org.springframework.data.util.TypeUtils;
 import org.springframework.data.projection.EntityProjectionIntrospector;
 import org.springframework.data.projection.TargetAware;
 import org.springframework.data.repository.Repository;
 import org.springframework.data.repository.core.RepositoryInformation;
 import org.springframework.data.repository.core.support.RepositoryFactoryBeanSupport;
 import org.springframework.data.repository.core.support.RepositoryFragment;
+import org.springframework.data.util.QTypeContributor;
+import org.springframework.data.util.TypeContributor;
+import org.springframework.data.util.TypeUtils;
 import org.springframework.lang.Nullable;
 import org.springframework.stereotype.Component;
 import org.springframework.util.Assert;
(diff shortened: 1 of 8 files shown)


src/main/java/org/springframework/data/util/QTypeContributor.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @author Christoph Strobl
+ * @since 4.1
+ */
+public class QTypeContributor {
+
+	private final static Log logger = LogFactory.getLog(QTypeContributor.class);
+	private static Function<ClassLoader, Class<?>> entityPathType = cacheOf(QTypeContributor::getEntityPathType);
+
+	public static void contributeEntityPath(Class<?> type, GenerationContext context, @Nullable ClassLoader classLoader) {
```
mp911de, line 39: How about making the entire code conditional based on the presence of `com.querydsl.core.types.EntityPath`?

src/main/java/org/springframework/data/repository/config/RepositoryRegistrationAotContribution.java
```diff
@@ -263,10 +264,12 @@ private void contributeRepositoryInfo(AotRepositoryContext repositoryContext, Ge
 		contribution.getRuntimeHints().reflection()
 				.registerType(repositoryInformation.getRepositoryInterface(),
 						hint -> hint.withMembers(MemberCategory.INVOKE_PUBLIC_METHODS))
-				.registerType(repositoryInformation.getRepositoryBaseClass(),
-						hint -> hint.withMembers(MemberCategory.INVOKE_DECLARED_CONSTRUCTORS, MemberCategory.INVOKE_PUBLIC_METHODS));
+				.registerType(repositoryInformation.getRepositoryBaseClass(), hint -> hint
+						.withMembers(MemberCategory.INVOKE_DECLARED_CONSTRUCTORS, MemberCategory.INVOKE_PUBLIC_METHODS));
 
 		TypeContributor.contribute(repositoryInformation.getDomainType(), contribution);
+		QTypeContributor.contributeEntityPath(repositoryInformation.getDomainType(), contribution,
```
mp911de, line 271: We should check for `QTypeContributor.QUERYDSL_PRESENT` or similar.

src/main/java/org/springframework/data/util/QTypeContributor.java
```diff
@@ -0,0 +1,109 @@
...
+		return in -> cache.computeIfAbsent(in, inFunction::apply);
+	}
+
+	/**
+	 * Returns the name of the query class for the given domain class.
+	 *
+	 * @param domainClass
+	 * @return
+	 */
+	private static String getQueryClassName(Class<?> domainClass) {
```
mp911de, line 89: We have a similar computation in our other Querydsl utils. It would make sense to reuse those parts.

src/main/java/org/springframework/data/util/QTypeContributor.java
```diff
@@ -0,0 +1,109 @@
...
+					logger.debug("Skipping Q type %s. Not an EntityPath.");
+				}
+			} catch (ClassNotFoundException e) {
+				throw new IllegalStateException("%s could not be loaded".formatted(queryClassName), e);
+			}
+		}
+	}
+
+	@Nullable
+	private static Class<?> getEntityPathType(ClassLoader classLoader) {
```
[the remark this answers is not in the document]
christophstrobl: I think we should use the same class loader that is used for AOT processing.

src/main/java/org/springframework/data/util/QTypeContributor.java
christophstrobl: true, but that created cycles between the packages.

src/main/java/org/springframework/data/util/QTypeContributor.java
christophstrobl: that's what happens here, just taking the `ClassLoader` into account.

PR #2746, opened 2022-12-07
Exclude local & anonymous classes from type inspection

christophstrobl:
_Local_ and _anonymous_ classes should not be picked up during reachable type collection, as those are not targets of the spring data reflection infrastructure. 
We also updated the type filter so that it now matches types present in the very same package and modified method visibility for module specific contribution checks.
Last we excluded _synthetic_ constructors from being inspected as those (in some cases) can contain references to anonymous classes as shown in the snippet below.

```java
public enum Outer {

    INSTANCE {
        @Override
        public Inner getInnerType() {
            return Inner.VALUE;
        }
    },

    public abstract Inner getInnerType();
}
```

```
synthetic Outer(java.lang.String arg0, int arg1, Outer$1 arg2)
         <localVar:index=0 , name=this , desc=LOuter;, sig=null, start=L0, end=L1>
         <localVar:index=1 , name=x0 , desc=Ljava/lang/String;, sig=null, start=L0, end=L1>
         <localVar:index=2 , name=x1 , desc=I, sig=null, start=L0, end=L1>
         <localVar:index=3 , name=x2 , desc=LOuter$1;, sig=null, start=L0, end=L1>
```

### DIFF
--- a/src/main/java/org/springframework/data/util/TypeCollector.java
+++ b/src/main/java/org/springframework/data/util/TypeCollector.java
@@ -48,16 +48,18 @@ public class TypeCollector {
 
 	private static final Log logger = LogFactory.getLog(TypeCollector.class);
 
-	static final Set<String> EXCLUDED_DOMAINS = new HashSet<>(Arrays.asList("java", "sun.", "jdk.", "reactor.",
-			"kotlinx.", "kotlin.", "org.springframework.core.", "org.springframework.data.mapping.",
-			"org.springframework.data.repository.", "org.springframework.boot.", "org.springframework.core."));
+	static final Set<String> EXCLUDED_DOMAINS = new HashSet<>(
+			Arrays.asList("java", "sun.", "jdk.", "reactor.", "kotlinx.", "kotlin.", "org.springframework.core.",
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/util/TypeCollector.java
```diff
@@ -125,6 +127,10 @@ Set<Type> visitConstructorsOfType(ResolvableType type) {
 		}
 		Set<Type> discoveredTypes = new LinkedHashSet<>();
 		for (Constructor<?> constructor : type.toClass().getDeclaredConstructors()) {
+
+			if (constructor.isSynthetic()) {
```
mp911de, line 131: Kotlin likes to generate synthetic constructors for data classes. We should make sure to not break stuff. Likely, the field-based inspection is going to catch such cases.

mp911de: See: https://github.com/spring-projects/spring-data-commons/blob/6a23723f07669e5d4031b3378b3af40e0d15eb82/src/main/java/org/springframework/data/mapping/model/KotlinClassGeneratingEntityInstantiator.java#L108

mp911de:

PR #2762, opened 2023-01-10
Use unique bean name for configuration post processors

christophstrobl:
This allows to pick up settings for more than one `@Enable...Repositories` configuration.


mp911de:

PR #2771, opened 2023-02-01
Query method parameters are now aware of aggregate reference type.

### LINKED ISSUE #2770: Query method parameter of type `Class` erroneously considered dynamic projection parameter
Consider a repository query method declaration like this:

```java
interface SomeRepository extends Repository<SomeAggregate, …> {

  @Query("some query using `type` as parameter")
  <T extends SomeAggregate> Collection<T> findOnlySpecificSubTypes(Class<T> type);
}
```

The query method intends to find only subtypes of `SomeAggregate` and the type being defined by the `type` parameter. Unfortunately, it is currently always considered a dynamic projection parameter, even if we are not dealing with a projection here.

The decision whether a parameter is considered a dynamic projection one is currently held in `Parameter` which as of now, unfortunately, doesn't know anything about the primary aggregate type. If it had that, we could back out of any projection considerations once we realize the `Class`' component type is tied to that.

A different approach could be to disregard a `Class` parameter annotated with `@Param` as projection parameter. The annotation currently requires a value to be set, which makes it a bit cumbersome to use, as you'd need to repeat the parameter name unconditionally. I am not sure whether making that optional would be a big issue.

odrotbohm:
To support the binding of Class parameters to queries for declared query methods, we have to be able to differentiate them from Class parameters that are supposed to represent projections. We can do that by relating the declared Class' element type to the aggregate root type, as a Class typed to that or any subtype of it will never trigger a projection by definition.

So far, the Parameter(s) abstraction was solely created from a query method's Method. We now changed that for QueryMethod to forward the aggregate type detected on the RepositoryMetadata and consider it during the detection of dynamic projection parameters.

As a mitigating measure, we now also support @Param on Class-typed parameters to explicitly mark them for query binding. This is primarily to be able to add this support to the 2.7

The changes are built in a way that modules extending that mechanism will continue to work as is but see deprecation warnings on methods and constructors involved. Adapting extending code to the new APIs will automatically enable the support for bindable Class parameters on query methods.

Fixes #2770.

### DIFF
--- a/src/main/java/org/springframework/data/repository/query/DefaultParameters.java
+++ b/src/main/java/org/springframework/data/repository/query/DefaultParameters.java
@@ -18,7 +18,7 @@ package org.springframework.data.repository.query;
 import java.lang.reflect.Method;
 import java.util.List;
 
-import org.springframework.core.MethodParameter;
+import org.springframework.data.util.TypeInformation;
 
 /**
  * Default implementation of {@link Parameters}.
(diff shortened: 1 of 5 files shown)


mp911de:
That's polished, merged and backported now.

Issue #1947, opened 2019-04-17
Add support for Kotlin Value Classes


spring-projects-issues:
**[Wyatt Smith](https://jira.spring.io/secure/ViewProfile.jspa?name=wyatt)** opened **[DATACMNS-1517](https://jira.spring.io/browse/DATACMNS-1517?redirect=false)** and commented

Here is an example with the bug: https://github.com/wyattjsmith1/SpringDataBug

When this runs, there is an `IndexOutOfBoundsException`. This is caused by kotlin's synthetic constructor. I believe the issues is that synthetic constructors aren't filtered before.

`buildPreferredConstructor` at `org.springframework.data.mapping.model.PreferredConstructorDiscoverer.Discoverers#discover`, but there is probably a better solution to this.

Changing AccountId in the data class to a `String` causes the application to work properly. Admittedly, inline classes are still experimental for Kotlin, but this might be worth investigating

---

**Issue Links:**
- [DATAGRAPH-1330](https://jira.spring.io/browse/DATAGRAPH-1330) Properties with type of Kotlin inline class has mangled name

7 votes, 10 watchers

matthewadams:
I'm seeing something similar to this on a brand new, greenfield project using Kotlin `1.5.10`, Spring Boot `2.5.1`, Spring Data `2021.0.1` (w/Spring Data Commons `2.5.1` & Spring Data MongoDB `3.2.1`).  I'm using immutable `data class`es for my `@Document`s and their value(s).  When I attempt to read a `Schedule` from mongo, I get the exception `org.springframework.data.mapping.MappingException: Parameter org.springframework.data.mapping.PreferredConstructor$Parameter@80a6e44c does not have a name!`.  The parameter in question is, indeed, the `DefaultConstructorMarker`.

The offending `data class` is :
```
sealed interface Availability {
    companion object {
        val CONVENTIONAL_8TO5_WITH_LUNCH_RECURRENCE: CronString = "0/30 8-12,13-17 * * MON-FRI"
        val CONVENTIONAL_8TO5_RECURRENCE: CronString = "0/30 8-17 * * MON-FRI"
        val CONVENTIONAL_9TO5_WITH_LUNCH_RECURRENCE: CronString = "0/30 9-12,13-17 * * MON-FRI"
        val CONVENTIONAL_9TO5_RECURRENCE: CronString = "0/30 9-17 * * MON-FRI"
[...]

PR #2773, opened 2023-02-03
Detect nullable marker parameters in constructor arguments. 

christophstrobl:
This **draft** seeks to introduce a `nullableMarker` flag to the `Parameter` abstraction. If set, it is safe skip and `null` the parameter value right away.

The `PreferredConstructorDiscoverer` for Kotlin will use this flag when detecting a `DefaultConstructorMarker` argument within the java constructor which is there (as the name indicates) to solely mark the constructor to use but does not map to any parameter from the source Kotlin type, resulting in an unresolvable parameter name/target property for the parameter.
This change will allow the `ClassGeneratingEntityInstantiator` to operate types using Kotlin inline classes.

I think it would also make sense to discuss if the usage of the newly introduced method makes sense in the `ParameterValueProvider` implementations like `PersistentEntityParameterValueProvider` to shortcut the value lookup by simply returning `null`

Closes: #1947

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>3.1.0-SNAPSHOT</version>
+	<version>3.1.x-GH-1947-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 	<description>Core Spring concepts underpinning every Spring Data module.</description>
--- a/src/main/java/org/springframework/data/mapping/Parameter.java
+++ b/src/main/java/org/springframework/data/mapping/Parameter.java
@@ -42,6 +42,7 @@ public class Parameter<T, P extends PersistentProperty<P>> {
 
 	private final Lazy<Boolean> enclosingClassCache;
 	private final Lazy<Boolean> hasSpelExpression;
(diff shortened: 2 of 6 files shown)


mp911de:
We have a few other flavors of non-domain arguments for constructors such as the outer class instance (for non-static inner classes) or nullable mask markers.

In all other places, we do not create a `Parameter` in the first place and I think we should follow the same pattern here as well.

christophstrobl:
So we'd end up with a signature that requires us to hand more arguments than we have parameters for, which is odd in first place.

mp911de:
The alternative to introducing all sorts of markers isn't a better variant. Kotlin nullability integer masks are spread over a few constructor integer parameters. For every 32 nullable parameters there's one additional integer bit mask.

christophstrobl:
have an example for that? I doubt it's targeting the same scenario.

christophstrobl:
since we do not have a stable API contract (see: [1947](https://github.com/spring-projects/spring-data-commons/issues/1947#issuecomment-1413947156)) other than the public constructor, and cannot safely invoke the method  with one parameter in the model missing skipping parameters seems to be a bad choice. 
Maybe there's a difference in the use case you mention above that would allow to ignore entire constructors, but not the parameter if we detect it.

PR #2787, opened 2023-03-06
Add support for `Scroll` API to allow Keyset- and Offset-based scrolling

### LINKED ISSUE #2151: Add support for keyset-based scrolling
**[Jens Schauder](https://jira.spring.io/secure/ViewProfile.jspa?name=schauder)** opened **[DATACMNS-1729](https://jira.spring.io/browse/DATACMNS-1729?redirect=false)** and commented

Value based pagination basically uses queries of the form: select n rows after the one with id value :id

On relational databases these kind of queries show better performance than the typical: select n rows after the first m especially when m is large.

It is likely that other stores benefit as well

---
No further details from [DATACMNS-1729](https://jira.spring.io/browse/DATACMNS-1729?redirect=false)

mp911de:
Tasks:

- [x] API spike
- [x] Implementation
- [x] Review naming
- [x] Documentation

Closes #2151


christophstrobl:
The name of `CursorRequest` comes with a lot of perception regarding its behaviour based on existing usage of the `Cursor` term. Keyset pagination itself is also a well established term for what we want to archive, so introducing yet another naming for the very same thing does not seem to be a good idea adding kognitive load for users. 
Paging does not necessarily need to return a `Page` as we've already seen with `Slice`, `Chunk` etc., so having something like a `Window` makes sense.
Maybe we'd be better off by naming it `Scrollable` and `ScrollRequest`, while sticking to `KeysetPageRequest` implementing a `ScrollRequest` having the method return a `Window`.

src/main/java/org/springframework/data/domain/Pageable.java
```diff
@@ -161,4 +161,20 @@ default Optional<Pageable> toOptional() {
...
+	 */
+	default CursorRequest toCursorRequest() {
+
+		if (isUnpaged()) {
+			return CursorRequest.uncursored();
+		}
+
+		return OffsetCursorRequest.ofSize(getPageSize(), getSort()).withOffset(getOffset());
+	}
+
```
rstoyanchev, line 179: The created `CursorRequest` points to page 0 and element offset 0 regardless of the page number. Shouldn't `Pageable.getOffset()` be passed into `ofSize`?

src/main/java/org/springframework/data/domain/CursorRequest.java
```diff
@@ -0,0 +1,126 @@
...
+	 */
+	Sort getSort();
+
+	/**
+	 * Creates a new cursor request with the {@link Sort}.
+	 *
+	 * @param sort the cursor result sort order.
+	 * @return a new {@link CursorRequest} with {@link Sort} applied.
+	 */
+	CursorRequest withSort(Sort sort);
```
christophstrobl, line 84: Why does a request need to know about the sort? I agree that there's value in the ability to change the window size with each request, but changing result order while scrolling through data does not make senes, does it?

mp911de: This was introduced to carry over the final sort from a query method that either defines a sort or accepts a `Sort` argument.

src/main/java/org/springframework/data/domain/KeysetScrollPosition.java
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	/**
+	 * Creates a new {@link KeysetScrollPosition} from a keyset and {@link Direction}.
+	 *
+	 * @param keys must not be {@literal null}.
+	 * @param direction must not be {@literal null}.
+	 * @return a new {@link KeysetScrollPosition} for the given keyset and {@link Direction}.
+	 */
+	public static KeysetScrollPosition of(Map<String, ?> keys, Direction direction) {
```
rstoyanchev, line 72: This is impossible to use currently since `Direction` is declared package private.

src/main/java/org/springframework/data/domain/Scroll.java
```diff
@@ -0,0 +1,167 @@
...
+ * A scroll of data consumed from an underlying query result. A scroll is similar to {@link Slice} in the sense that it
+ * contains a subset of the actual query results for easier consumption of large result sets. The scroll is less
+ * opinionated about the actual data retrieval, whether the query has used index/offset, keyset-based pagination or
+ * cursor resume tokens.
+ *
+ * @author Mark Paluch
+ * @since 3.1
+ * @see ScrollPosition
+ */
+public interface Scroll<T> extends Streamable<T> {
```
christophstrobl, line 35: I think `Window` was the better therm here. It aligned nicely with `ScrollPosition`.

src/main/asciidoc/repositories-scrolling.adoc
```diff
@@ -0,0 +1,83 @@
...
+Scroll<User> users = repository.findFirst10ByLastnameOrderByFirstname("Doe", OffsetScrollPosition.initial());
+
+do {
+
+  for (User u : users) {
+    // consume the user
+  }
+
+  // obtain the next Scroll
+  users = repository.findFirst10ByLastnameOrderByFirstname("Doe", users.lastScrollPosition());
```
christophstrobl, line 37: When reading this, I had a hard time to find out what the `ScrollPosition` actually does. Is it including, like give me everything starting (inclusive) at a certain position. Or is it excluding, like return the data and start after the position. 
My point here is, should we call it `lastScrollPosition` or would it better be `nextScrollPosition` which is also not quite accurate. Does it maybe make sense to have a scroll position that has `startAt` and `startAfter`?
Does this read any better `ScrollPosition.after(users.lastScrollPosition())` or is it even more confusing?

christophstrobl:
Merged to _main_ development line.

PR #2825, opened 2023-04-30
Fix `Sort` javadoc

scordio:


schauder:
Thanks for spotting and fixing that.

PR #2836, opened 2023-05-31
Introduce `Limit` type

christophstrobl:
`Limit` can be used as `Parameter` within repository query methods to define the maximum number of results to be returned. 

```java
List<Person> findByFirstnameLike(String firstname, Limit limit);

List<Person> findByFirstnameLike(String firstname, Sort orderBy, Limit limit);

Window<User> findByFirstnameLikeOrderByLastname(ScrollPosition pageable, Limit limit);
```

The default `ParameterAccessor` implementation will pick up `Limit` and `Sort` to assemble a `Pageable` out of the provided values if present to provide out of the box support for modules using `ParameterAcessor#getPageable()` to apply order and limit.

`Limit#max` is aligned to `Page.size`, both using `int`. We might want to consider `long` for `Limit` and error if the actual value exceeds the `Page` one when assembling the `Pageable`.

### DIFF
--- a/src/main/asciidoc/repositories-paging-sorting.adoc
+++ b/src/main/asciidoc/repositories-paging-sorting.adoc
@@ -1,12 +1,12 @@
 [[repositories.special-parameters]]
-=== Paging, Iterating Large Results, Sorting
+=== Paging, Iterating Large Results, Sorting & Limiting
 
 To handle parameters in your query, define method parameters as already seen in the preceding examples.
-Besides that, the infrastructure recognizes certain specific types like `Pageable` and `Sort`, to apply pagination and sorting to your queries dynamically.
+Besides that, the infrastructure recognizes certain specific types like `Pageable`, `Sort` and `Limit`, to apply pagination, sorting and limiting to your queries dynamically.
 The following example demonstrates these features:
 
 ifdef::feature-scroll[]
-.Using `Pageable`, `Slice`, `ScrollPosition`, and `Sort` in query methods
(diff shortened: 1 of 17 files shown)


src/main/asciidoc/repositories-paging-sorting.adoc
```diff
@@ -57,6 +61,28 @@ Rather, it restricts the query to look up only the given range of entities.
...
+| `findBy...(Pageable page, Sort sort)`
+| `Pageable` already defines `Sort`
+
+| `Pageable` & `Limit`
+| `findBy...(Pageable page, Limit limit)`
+| `Pageable` already defines a limit.
+
+|===
+
+The `Top` keyword used to limit results can be used to along with `Pageable` whereas `Top` defines the total maximum of results, whereas the Pageable parameter may reduce this this number.
```
odrotbohm, line 83: Double "this this"?

src/main/java/org/springframework/data/repository/query/QueryMethod.java
```diff
@@ -290,6 +264,41 @@ public String toString() {
 		return method.toString();
 	}
 
+	public void validate() {
```
odrotbohm, line 267: Needs to be public?

odrotbohm (commented):
A couple of minor findings.

odrotbohm (commented):
A couple of minor findings.

mp911de (changes requested):
Left some comments, seconding what Ollie said.

mp911de:

Issue #2840, opened 2023-06-06
KeysetScrollPosition.of(Collections.emptyMap(), Direction.BACKWARD) retain incorrect direction


quaff:
https://github.com/spring-projects/spring-data-commons/blob/bfcb2ffeae4a9792c8f9abc349b5bcf82ad3fb99/src/main/java/org/springframework/data/domain/KeysetScrollPosition.java#L68-L76

Parameter `direction` should not be ignored if keys is empty.

quaff:
Superseded by https://github.com/spring-projects/spring-data-commons/pull/2841

PR #2841, opened 2023-06-06
Parameter direction should not be ignored if keys is empty

quaff:
After this commit, KeysetScrollPosition.of(Collections.emptyMap(), Direction.BACKWARD) will retain correct direction

Closes GH-2840


quaff:
It seems an obvious fix, could you take a look? @mp911de

mp911de:
Thank your for your pull request. We took a slightly different approach and it made sense to implement that change directly instead of polishing up the pull request.

PR #2843, opened 2023-06-06
Introduce convenient method Window::positionForNext

quaff:


quaff:
The new convenient method is useful for https://github.com/spring-projects/spring-data-commons/commit/05d68a0eab0e4442dd30b843cdbfaef02350b2b0, could you review this? @mp911de

quaff:
> We initially had a similar design where we could obtain positions for next/previous windows from the `Window`. This resulted in an ambiguous design and therefore we decided to not have `next` methods that can produce a next window.

Could you explain the ambiguity?

mp911de:
When having a `next…` method, then this design asks for a `previous…` method as well. The window itself isn't capable of fetching data, hence it must return a `ScrollPosition`. When bringing that design in the context of directional scrolling, then backward `next` means actually previous and vice versa.

quaff:
> When having a `next…` method, then this design asks for a `previous…` method as well. The window itself isn't capable of fetching data, hence it must return a `ScrollPosition`. When bringing that design in the context of directional scrolling, then backward `next` means actually previous and vice versa.

I don't think we need `previous…`, just like there is `Window::hasNext` but no `Window::hasPrevious`, we are expecting next not previous window for scrolling at most case.

Issue #2857, opened 2023-06-16
WindowIterator should emit items in order when scrolling backwards.


christophstrobl:
Scrolling backwards, `WindowIterator` should emit items in order eg. 10, 9, 8, 7,...
Currently _backward_ scrolling is only considered on a `Window` basis, but items within the `Window` are ordered forward.
Using a `limit` of eg. 3 would therefore lead to something like 8, 9, 10, 5, 6, 7...

PR #2858, opened 2023-06-16
Emit items in order when scrolling backwards using WindowIterator.

christophstrobl:
Closes: #2857


mp911de:

Issue #2856, opened 2023-06-16
Introduce argument resolver for `OffsetScrollPosition`


quaff:
Just like `org.springframework.data.web.PageableHandlerMethodArgumentResolver` did for `Pageable`

mp911de:
We do not have a good way to capture the keyset type. For offset-based positions, we could provide something. Feel free to submit a pull request. We should provide argument resolves for the servlet- and the reactive stack.

quaff:
I will try `OffsetScrollPosition` first.

quaff:
For `KeysetScrollPosition`:

Given
```java
@GetMapping("/users")
public Window<User> scroll(@SortDefault(sort = "id", direction = ASC) Sort sort,
		@RequestParam(defaultValue = "10") int limit, KeysetScrollPosition position) {
	Specification<User> spec = (root, query, cb) -> null;
	return this.userRepository.findBy(spec, q -> q.limit(limit).sortBy(sort).scroll(position));
}
```
Use implicit `id` as sort:
```
/users -> ScrollPosition.forward(Map.of())
/users?id=10 -> ScrollPosition.forward(Map.of("id", 10))
/users?direction=backward -> ScrollPosition.backward(Map.of())
/users?direction=backward&id=10 -> ScrollPosition.backward(Map.of("id", 10))
````

Use explicit `username` as sort:
```
/users?sort=username -> ScrollPosition.forward(Map.of())
/users?sort=username&username=test -> ScrollPosition.forward(Map.of("username", "test"))
/users?sort=username&direction=backward -> ScrollPosition.backward(Map.of())
/users?sort=username&direction=backward&username=test -> ScrollPosition.backward(Map.of("username", "test"))
```

Use explicit  `foo` and `bar`  combination as sort:
```
/users?sort=foo,bar -> ScrollPosition.forward(Map.of())
/users?sort=foo,bar&foo=foo&bar=bar -> ScrollPosition.forward(Map.of("foo", "foo", "bar", "bar"))
/users?sort=foo,bar&direction=backward -> ScrollPosition.backward(Map.of())
/users?sort=foo,bar&direction=backward&foo=foo&bar=bar -> ScrollPosition.backward(Map.of("foo", "foo", "bar", "bar"))
```
WDYT? @mp911de

PR #2861, opened 2023-06-16
Introduce argument resolver for `OffsetScrollPosition`

quaff:
See GH-2856


quaff:
Should the class names `OffsetScrollPosition*` be simplified as `Offset*`?

mp911de:
> Should the class names OffsetScrollPosition* be simplified as Offset*?

It is fine to have fully-qualified class names as we want to be explicit about the type of arguments we want to resolve.

mp911de:

PR #2865, opened 2023-06-18
Use `Pageable.unpaged(sort)` for sorted unpaged pageable

quaff:
Fix
```java
java.lang.UnsupportedOperationException
	at org.springframework.data.domain.Unpaged.getPageNumber(Unpaged.java)
	at org.springframework.data.web.ReactivePageableHandlerMethodArgumentResolver.resolveArgumentValue(ReactivePageableHandlerMethodArgumentResolver.java)
```


themadprofessor:
Would it be possible to handle sorted unpaged pagables as well?

quaff:
> Would it be possible to handle sorted unpaged pagables as well?

Thanks for your suggestion, I've updated the PR.

odrotbohm:
That's been polished and merged. Thanks!

PR #2866, opened 2023-06-22
Add support for Kotlin Value Classes

### LINKED ISSUE #2806: Add support for Kotlin Value Classes
Right now, we do not detect constructors and constructor arguments for classes that use Kotlin value classes because they do not use typically a defaulting mask.

Also, a lot from the inline classes mashup happens in the sense of rewritten property names, rewritten copy method name. Additionally, nullability infers boxing of value classes so for nullable  components, Kotlin boxes values in inline classes and then unboxes these:

```
public final class KotlinClassGeneratingEntityInstantiatorUnitTests$WithMyNullableInlineClass {
   @NotNull
   private final String id;

   private KotlinClassGeneratingEntityInstantiatorUnitTests$WithMyNullableInlineClass(String id) {
      this.id = id;
   }

   // $FF: synthetic method
   public KotlinClassGeneratingEntityInstantiatorUnitTests$WithMyNullableInlineClass(KotlinClassGeneratingEntityInstantiatorUnitTests.MyNullableInlineClass var1, int var2, DefaultConstructorMarker var3) {
      if ((var2 & 1) != 0) {
         var1 = MyNullableInlineClass.box-impl(MyNullableInlineClass.constructor-impl("foo"));
      }

      this(var1.unbox-impl(), (DefaultConstructorMarker)null);
   }

   @NotNull
   public final String getId_MYsf4WQ/* $FF was: getId-MYsf4WQ*/() {
      return this.id;
   }

   @NotNull
   public final String component1_MYsf4WQ/* $FF was: component1-MYsf4WQ*/() {
      return this.id;
   }

   @NotNull
   public final KotlinClassGeneratingEntityInstantiatorUnitTests$WithMyNullableInlineClass copy_jv4avjM/* $FF was: copy-jv4avjM*/(@NotNull String id) {
      return new KotlinClassGeneratingEntityInstantiatorUnitTests$WithMyNullableInlineClass(id, (DefaultConstructorMarker)null);
   }

   // $FF: synthetic method
   public static KotlinClassGeneratingEntityInstantiatorUnitTests$WithMyNullableInlineClass copy_jv4avjM$default/* $FF was: copy-jv4avjM$default*/(KotlinClassGeneratingEntityInstantiatorUnitTests$WithMyNullableInlineClass var0, KotlinClassGeneratingEntityInstantiatorUnitTests.MyNullableInlineClass var1, int var2, Object var3) {
      if ((var2 & 1) != 0) {
         var1 = MyNullableInlineClass.box-impl(var0.id);
      }

      return var0.copy-jv4avjM(var1.unbox-impl());
   }

   @NotNull
   public String toString() {
      // …
   }

   public int hashCode() {
      // …
   }

   public boolean equals(@Nullable Object other) {
      // …
   }

   // $FF: synthetic method
   public KotlinClassGeneratingEntityInstantiatorUnitTests$WithMyNullableInlineClass(String id, DefaultConstructorMarker $constructor_marker) {
      this(id);
   }
}
```

mp911de:
We now support instantiation and mutation of types using Kotlin Value (`@JvmInline`) classes:

```kotlin
@JvmInline
value class MyValueClass(val id: String)

data class WithMyValueClass(val id: MyValueClass) { }
```

```java
class WithMyValueClass {
  private final String id;

  WithMyValueClass(String id){…}
  // getters, setter, copy
}
```

Value classes are typically flattened into their value type (component type) if the type at the declaration site does not change. Nullable value classes will be boxed.

```kotlin
@JvmInline
value class MyNullableValue(val id: String?)

data class WithMyNullableValue(val id: MyNullableValue?) { }
```

```java
class MyNullableValue {
  private final MyNullableValue id;

  WithMyValueClass(MyNullableValue id){…}
  // getters, setter, copy
}
```

Boxed values result in being recognized as entities.

The following components are affected by this change:

* `KotlinBeanInfoFactory`
* Reflective and class-generating `EntityInstantiator`
* Reflective and class-generating `PersistentPropertyAccessorFactory`
  * Kotlin reflection (or the inline-class compilation) suffers from a bug preventing the use of copy methods: https://youtrack.jetbrains.com/issue/KT-57357

Closes #2806


src/main/java/org/springframework/data/util/KotlinBeanInfoFactory.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * {@link BeanInfoFactory} specific to Kotlin types using Kotlin reflection to determine bean properties.
+ *
+ * @author Mark Paluch
+ * @since 3.2
+ * @see JvmClassMappingKt
+ * @see ReflectJvmMapping
+ */
+public class KotlinBeanInfoFactory implements BeanInfoFactory, Ordered {
```
christophstrobl, line 46: @sdeleuze what's your take on moving this one to core framework?

sdeleuze: I have in my todo to have a broader look to value class support in Framework, I will share my feedback on this specific point after and comment here.

odrotbohm:
I wonder if more of the underlying Kotlin class analysis code should / would need to move to Spring Framework, so that the standard framework binding subsystem was enabled to populate these types as well?

christophstrobl:
merged to _main_ development line.

src/main/java/org/springframework/data/mapping/model/KotlinValueUtils.java
```diff
@@ -0,0 +1,397 @@
...
+	public static boolean hasValueClassProperty(Class<?> type) {
+
+		if (!KotlinDetector.isKotlinType(type)) {
+			return false;
+		}
+
+		KClass<?> kotlinClass = JvmClassMappingKt.getKotlinClass(type);
+
+		for (KCallable<?> member : kotlinClass.getMembers()) {
+			if (member instanceof KProperty<?> kp && isValueClass(kp.getReturnType())) {
```
udalov, line 71: @mp911de Sorry for trespassing, but I believe this might work incorrectly in case there are unrelated properties in the class. For example:
```
@JvmInline
value class Z(val value: String)

data class D(val x: Int) {
    val unrelated1 = Z("1")

    val Any.unrelated2: Z get() = Z("2")
}
```

`hasValueClassProperty` will return true for the class `D`, but it probably should be false (if I understood call sites correctly).

I suppose you're interested in finding only if any property in the primary constructor has a value class type? For that you need to take only non-extension properties (i.e. [`KClass.memberProperties`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect.full/member-properties.html)) whose names are equal to the parameters of the primary constructor.

src/main/java/org/springframework/data/mapping/model/KotlinValueUtils.java
```diff
@@ -0,0 +1,397 @@
...
+				}
+
+				return expandedUnderlyingType;
+			}
+
+			static List<KProperty<?>> getProperties(KClass<?> kClass) {
+
+				if (kClass.isValue()) {
+
+					for (KCallable<?> member : kClass.getMembers()) {
```
udalov, line 189: @mp911de Similarly to the above, this should use `memberProperties`, otherwise it'll fail on something like this:
```
@JvmInline
value class Z(val value: String) {
    val String.unrelated: String get() = this
}
```

src/main/java/org/springframework/data/mapping/model/KotlinValueUtils.java
```diff
@@ -0,0 +1,397 @@
...
+
+		private ValueBoxing(BoxingRules rules, KType type, KClass<?> kClass, boolean optional) {
+
+			KFunction<?> wrapperConstructor = null;
+			ValueBoxing next = null;
+			boolean applyBoxing;
+
+			if (kClass.isValue()) {
+
+				wrapperConstructor = kClass.getConstructors().iterator().next();
```
udalov, line 243: @mp911de There's no guarantee that the first constructor is the primary one, nor that it has any parameters, so this can fail on something like this:
```
@JvmInline
value class Z(val value: String) {
    constructor() : this("Fail")
}
```
I recommend to use [`primaryConstructor`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect.full/primary-constructor.html) instead.

src/main/java/org/springframework/data/mapping/model/KotlinValueUtils.java
```diff
@@ -0,0 +1,397 @@
...
+				KParameter nested = wrapperConstructor.getParameters().get(0);
+				KType nestedType = nested.getType();
+
+				applyBoxing = rules.shouldApplyBoxing(type, optional, nested);
+
+				KClass<?> nestedClass;
+
+				// bound flattening
```
udalov, line 251: @mp911de You can use [`jvmErasure`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect.jvm/jvm-erasure.html) for this.

PR #2874, opened 2023-07-12
Introduce support to pass-thru TemporalAccessor auditing values

### LINKED ISSUE #2719: Add support to pass-thru `TemporalAccessor` auditing values
This is incorrect, it would seem that java.time.Instant is also supported, sorry if this belongs in the commons repo. Note, I was trying to use OffsetDateTime for this.

  businessDivisionId=a
  createdOn=<null>
  id=0183ae78-9359-7422-a7b7-471f29af8980
  lastModifiedOn=<null>
  version=0
]; Supported types are [org.joda.time.DateTime, org.joda.time.LocalDateTime, java.util.Date, java.lang.Long, long]

see this ticket, but especially this comment.

https://github.com/spring-projects/spring-data-commons/issues/880#issuecomment-1073194040

Most databases support "with timezone", javascript Date only seems to support offset, and so it is useful in cases to be able to communicate that and store it in the database.

mp911de:
We now allow passing-thru `TemporalAccessor` auditing values, bypassing conversion if the target value type matches the value provided from e.g. `DateTimeProvider`.

Refined the error messages and listing all commonly supported types for which we provide converters.

Closes #2719

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>3.2.0-SNAPSHOT</version>
+	<version>3.2.0-GH-2719-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 	<description>Core Spring concepts underpinning every Spring Data module.</description>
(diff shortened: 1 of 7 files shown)


schauder:
Thanks, that's merged.

PR #2882, opened 2023-07-18
Refine parameter and field names to align with `PageRequest` property names

thachlp:
We mix using page with pageNumber, size with pageSize and sometimes, it makes me confused. And I think pageNumber and pageSize are more clearer. Feel free to close this if there is a reason or we can not change it.

### DIFF
--- a/src/main/java/org/springframework/data/domain/AbstractPageRequest.java
+++ b/src/main/java/org/springframework/data/domain/AbstractPageRequest.java
@@ -24,49 +24,50 @@ import java.io.Serializable;
  * @author Oliver Gierke
  * @author Alex Bondarev
  * @author Johannes Englmeier
+ * @author Thach Le
  */
 public abstract class AbstractPageRequest implements Pageable, Serializable {
 
 	private static final long serialVersionUID = 1232825578694716871L;
 
-	private final int page;
-	private final int size;
+	private final int pageNumber;
+	private final int pageSize;
 
 	/**
-	 * Creates a new {@link AbstractPageRequest}. Pages are zero indexed, thus providing 0 for {@code page} will return
-	 * the first page.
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/domain/AbstractPageRequest.java
```diff
@@ -24,49 +24,50 @@
...
+	 * Creates a new {@link AbstractPageRequest}. Pages are zero indexed, thus providing 0 for {@code pageNumber} will return
+	 * the first pageNumber.
 	 *
-	 * @param page must not be less than zero.
-	 * @param size must not be less than one.
+	 * @param pageNumber must not be less than zero.
+	 * @param pageSize must not be less than one.
 	 */
-	public AbstractPageRequest(int page, int size) {
+	protected AbstractPageRequest(int pageNumber, int pageSize) {
```
mp911de, line 43: We should not change visibility. I'm going to revert this change upon merge.

src/main/java/org/springframework/data/domain/PageRequest.java
```diff
@@ -36,13 +37,13 @@ public class PageRequest extends AbstractPageRequest {
 	/**
 	 * Creates a new {@link PageRequest} with sort parameters applied.
 	 *
-	 * @param page zero-based page index, must not be negative.
-	 * @param size the size of the page to be returned, must be greater than 0.
+	 * @param pageNumber zero-based pageNumber index, must not be negative.
```
mp911de, line 40: Whoops, seems too much was changed by the refactoring. I'll align these upon merge.

mp911de:

jeffryang24:
Hi @mp911de, I think we need to update the release note for Spring Data 2022.0 at https://github.com/spring-projects/spring-data-commons/wiki/Spring-Data-2022.0-(Turing)-Release-Notes since this commit is considered a breaking change, especially for a downstream service which is consumed by many user. Wdyt? :bow:

PR #2883, opened 2023-07-19
Initial Docs Build

rwinch:


mp911de:
What's the background of having the docs build on a branch? Is there a way we can have this in our `main` branch?

rwinch:
@mp911de The `docs-build` branch is a detached branch similar to and for the same reasons that `gh-pages` tends to be detached. The `docs-build` branch builds the documentation for a set of branches (including but not exclusively `main`) and tags. While it will functionally work, placing this logic in `main` does not make much sense because it builds more than the documentation for `main` and should not be in tags that will come from `main`.

gregturn:
Merged to `docs-build`.

PR #2903, opened 2023-08-16
Added supplementary comment in AbstractAggregateRoot.

shin-mallang:
The registerEvent method works for either Spring Data repository's save or delete methods, but the comments only describe save.
I think it should be explained that it also works for the delete method, like in the EventPublishingMethodInterceptor class.


mp911de:

PR #2916, opened 2023-08-28
Introduce `PersistentProperty.isReadable`

### LINKED ISSUE #2915: Introduce `PersistentProperty.isReadable`
Based on spring-projects/spring-data-mongodb#4485, we should introduce a `isReadable` flag for easier downstream usage of truly immutable properties. The immutable flag just expresses immutability on the instance but requires checks for withers and Kotlin copy methods for creating a new instance. That isn't handy for values of Kotlin data classes (properties of a Kotlin data class that cannot be set via `copy`).

mp911de:
`isReadable` reports whether a property can be read through `PersistentPropertyAccessor`, by either using property access through setters, a wither, Kotlin Copy method or by accessing the field directly.

Closes #2915

### DIFF
--- a/src/main/java/org/springframework/data/mapping/model/AbstractPersistentProperty.java
+++ b/src/main/java/org/springframework/data/mapping/model/AbstractPersistentProperty.java
@@ -27,6 +27,7 @@ import java.util.stream.Collectors;
 import org.springframework.data.mapping.Association;
 import org.springframework.data.mapping.PersistentEntity;
 import org.springframework.data.mapping.PersistentProperty;
+import org.springframework.data.util.KotlinReflectionUtils;
 import org.springframework.data.util.Lazy;
 import org.springframework.data.util.ReflectionUtils;
 import org.springframework.data.util.TypeInformation;
@@ -70,6 +71,7 @@ public abstract class AbstractPersistentProperty<P extends PersistentProperty<P>
 	private final Method setter;
 	private final Field field;
 	private final Method wither;
+	private final Lazy<Boolean> readable;
 	private final boolean immutable;
 
 	public AbstractPersistentProperty(Property property, PersistentEntity<?, P> owner,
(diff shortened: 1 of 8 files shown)


src/main/java/org/springframework/data/mapping/model/AbstractPersistentProperty.java
```diff
@@ -103,11 +105,27 @@ public AbstractPersistentProperty(Property property, PersistentEntity<?, P> owne
...
 
-		if (setter == null && (field == null || Modifier.isFinal(field.getModifiers()))) {
-			this.immutable = true;
-		} else {
-			this.immutable = false;
-		}
+		this.immutable = setter == null && (field == null || Modifier.isFinal(field.getModifiers()));
+		this.readable = Lazy.of(() -> {
+
+			if (setter != null) {
```
odrotbohm, line 111: How does the presence of a setter make a property readable? 🤔

src/main/java/org/springframework/data/mapping/model/AbstractPersistentProperty.java
```diff
@@ -103,11 +105,27 @@ public AbstractPersistentProperty(Property property, PersistentEntity<?, P> owne
...
+				return true;
+			}
+
+			if (wither != null) {
```
odrotbohm, line 115: Same for a wither?

odrotbohm (commented):
The lack of the method is probably an oversight, as I generally like symmetry in method declarations. Some inline questions regarding the documentation though.

src/main/java/org/springframework/data/mapping/model/AbstractPersistentProperty.java
mp911de: If there's no setter, then there's no way to set the property. While reading from the database, we need to put the value somewhere.

odrotbohm: I am sorry, I totally misunderstood what `….isReadable()` is even supposed to mean and totally got it backwards. I guess it's been a while. 😅

PR #2920, opened 2023-08-31
Remove unnecessary static keywords from methods in `EventPublishingRepositoryProxyPostProcessor`

shin-mallang:
I see no reason why `isEventPublishingMethod()`, `isSaveMethod()`, and `isDeleteMethod()` should be static methods.
They are only used within the `invoke()` of the `EventPublishingMethodInterceptor`.


mp911de:
These are static because they do not access state from the instance.

shin-mallang:
I see that was your intention.
I apologize for this PR.😥

PR #2927, opened 2023-09-13
Change Collection to Iterable for event publishing

quaff:
`repository.saveAll(Window<?>)` will be handled properly after this commit.

Fixes https://github.com/spring-projects/spring-data-jpa/issues/3153

### DIFF
--- a/src/main/java/org/springframework/data/repository/core/support/EventPublishingRepositoryProxyPostProcessor.java
+++ b/src/main/java/org/springframework/data/repository/core/support/EventPublishingRepositoryProxyPostProcessor.java
@@ -47,6 +47,7 @@ import org.springframework.util.ReflectionUtils;
  * @author Christoph Strobl
  * @author Yuki Yoshida
  * @author Réda Housni Alaoui
+ * @author Yanming Zhou
  * @since 1.13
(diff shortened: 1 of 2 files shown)


mp911de:
We deliberately chose to not consider `Iterable` types as domain types commonly implement `Iterable` or `Streamable` to hold inner values (`Order implements Iterable<OrderPosition>`). Therefore, we do not want to dive into such objects but consider events of the aggregate root.

quaff:
> We deliberately chose to not consider `Iterable` types as domain types commonly implement `Iterable` or `Streamable` to hold inner values (`Order implements Iterable<OrderPosition>`). Therefore, we do not want to dive into such objects but consider events of the aggregate root.

OK, it makes sense, and we should treat `Window` as `Collection` to fix this, I don't think any domain type will implements `Window`.

mp911de:
We have the same case with `Page` where the page type implements `Iterable` & `Streamable`.

quaff:
> We have the same case with `Page` where the page type implements `Iterable` & `Streamable`.

I've updated https://github.com/spring-projects/spring-data-commons/pull/2931 to include `Page`.

PR #2930, opened 2023-09-13
Make `Lazy.orElseGet(Supplier)` public

gregturn:


mp911de:

PR #2944, opened 2023-09-27
Introduce `Lock` utility.

mp911de:
The Lock utility is an adapter for Java's Lock and ReadWrite Lock types providing an easier to consume programming model (callback-style and try-with-resources).

This effort is inspired by repetitive `lock.lock(), try{} finally{lock.unlock();}` usage where it is too simple to miss an unlock call.

### DIFF
--- a/src/main/java/org/springframework/data/util/Lock.java
+++ b/src/main/java/org/springframework/data/util/Lock.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


mp911de:
See spring-projects/spring-data-mongodb#4431 for more usage examples.

src/main/java/org/springframework/data/util/Lock.java
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * @author Mark Paluch
+ * @since 3.2
+ */
+public interface Lock {
+
+	/**
+	 * Create a new {@link Lock} adapter for the given {@link java.util.concurrent.locks.Lock delegate}.
+	 *
+	 * @param delegate must not be {@literal  null}.
```
christophstrobl, line 52: there are two spaces in `{@literal  null}`

src/test/java/org/springframework/data/util/LockUnitTests.java
```diff
@@ -0,0 +1,116 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.springframework.data.util.Lock.AcquiredLock;
+
+/**
+ * Unit tests for {@link Lock}.
+ *
+ * @author Mark Paluch
+ */
+class LockUnitTests {
```
christophstrobl, line 31: should we have tests for the case when a the lock is acquired on the `delegate` before using one of the `execute` methods expecting them to increase the hold count?

christophstrobl: I think we should also test for exceptions that occur while holding the lock.
```java
lock.executeWithoutResult(() -> { throw ... });
```

mp911de:
Done. Rebased because of changes on `main`.

PR #2945, opened 2023-09-28
Proposal to fix #2531 adding explode modifier `*` to `sort` attribute

### LINKED ISSUE #2531: HateoasSortHandlerMethodArgumentResolver should append * to the 'sort' parameter template variable.
The [Spring docs](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#core.web.basic) (see Table 1) specify that *multiple* sort parameters should be present in the request params for Pageable instances when sorting by multiple fields.  (e.g. sort=field1,asc&sort=field2,desc).   

Spring HATEOAS via the data commons HateoasSortHandlerMethodArgumentResolver class generates a URI template variable for the 'sort' param *without* the URI template [explode modifier (*)](https://datatracker.ietf.org/doc/html/rfc6570#section-2.4.2).

```
http://localhost:8080/api/clients{?page,size,sort,projection}
```

For client side libraries performing URI expansion based on rfc6570, in conditions where there are multiple sort parameters, the template provided by Spring without the modifier will generate URLs of the following format:

```js
// using https://www.npmjs.com/package/uri-template based on rfc6570
const templ = uriTemplate.parse("http://localhost:8080/api/clients{?page,size,sort,projection}");
const expanded = templ.expand({sort : ["id,asc", "lastModifiedDate,desc"]});

// BAD -> http://localhost:8080/api/clients?sort=id%2Casc,lastModifiedDate%2Cdesc
```

This URL isn't in the correct format outlined in the [Spring docs](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#core.web.basic) (see Table 1), and a format not expected by the handlers resulting in incorrect parsing (results in two order by's, but direction is *DESC* for both).     

The explode modifier added to sort would allow rfc6570 compliant libraries to generate the correct URLs expected by Spring and allow correct sorting criteria to be generated:

```js
const templ = uriTemplate.parse("http://localhost:8080/api/clients{?page,size,sort*,projection}");
const expanded = templ.expand({sort : ["id,asc", "lastModifiedDate,desc"]});

// GOOD -> http://localhost:8080/api/clients?sort=id%2Casc&sort=lastModifiedDate%2Cdesc
```

There is an existing related issue #1242, which deals with multiple values and generated query method templates... which i can see needs addressing deeper in the HATEOAS codebase, but to fix sorting properly by anything Pageable, it seems like a simple fix would be to append "*" to the sort param generated by this class.  

[org.springframework.data.web.HateoasSortHandlerMethodArgumentResolver#getSortTemplateVariables](https://github.com/spring-projects/spring-data-commons/blob/da5df3495e20af92d517caefe1900cbe089b4658/src/main/java/org/springframework/data/web/HateoasSortHandlerMethodArgumentResolver.java#L50)

motofix:
Adding explode modifier `*` to `sort` attribute, making client libraries like Ketting to behave correctly and generate request compliant with Spring

Proposal to fix #2531

### DIFF
--- a/src/main/java/org/springframework/data/web/HateoasSortHandlerMethodArgumentResolver.java
+++ b/src/main/java/org/springframework/data/web/HateoasSortHandlerMethodArgumentResolver.java
@@ -57,11 +57,9 @@ public class HateoasSortHandlerMethodArgumentResolver extends SortHandlerMethodA
 			return TemplateVariables.NONE;
 		}
 
-		String sortParameterExplode = sortParameter + '*';
-
-		String description = String.format("pagination.%s.description", sortParameterExplode);
+		String description = String.format("pagination.%s.description", sortParameter);
 		TemplateVariable.VariableType type = append ? REQUEST_PARAM_CONTINUED : REQUEST_PARAM;
-		return new TemplateVariables(new TemplateVariable(sortParameterExplode, type, description));
+		return new TemplateVariables(new TemplateVariable(sortParameter, type, description).composite());
 	}
 
 	@Override
--- a/src/main/java/org/springframework/data/web/HateoasSortHandlerMethodArgumentResolver.java
+++ b/src/main/java/org/springframework/data/web/HateoasSortHandlerMethodArgumentResolver.java
@@ -35,6 +35,7 @@ import org.springframework.web.util.UriComponentsBuilder;
  * @author Oliver Gierke
  * @author Thomas Darimont
  * @author Nick Williams
+ * @author Julien Béti
  */
 public class HateoasSortHandlerMethodArgumentResolver extends SortHandlerMethodArgumentResolver
 		implements UriComponentsContributor {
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/web/HateoasSortHandlerMethodArgumentResolver.java
```diff
@@ -56,9 +57,11 @@ public TemplateVariables getSortTemplateVariables(MethodParameter parameter, Uri
 			return TemplateVariables.NONE;
 		}
 
-		String description = String.format("pagination.%s.description", sortParameter);
+		String sortParameterExplode = sortParameter + '*';
```
odrotbohm, line 60: I think we should rather call `TemplateVariable.composite()` on the instance created below.

odrotbohm (changes requested):

src/main/java/org/springframework/data/web/HateoasSortHandlerMethodArgumentResolver.java
motofix: Thank you

motofix:
Hi, any news on this ?

odrotbohm:
Sorry for the delay and thanks for bumping this. I'll have a loot ASAP, this time for sure, I promise.

odrotbohm:
This is polished and back-ported to the 3.1.x maintenance branch. Thanks again for your patience!

PR #2959, opened 2023-10-18
Add overload to ReactiveSortingRepository accepting a Limit parameter.

### LINKED ISSUE #2923: Add overload to ReactiveSortingRepository accepting a `Limit` parameter.
```
Flux<T> findAll(Sort sort, Limit limit);
```

hwan33:
Related issue: #2923

### Motivation
`Flux<T> findAll(Sort sort, Limit limit);`

- It's good to add `findAll(sort, limit)` on `ReactiveSortingRepository`

### Modification
- Add `findAll(sort, limit)` on `ReactiveSortingRepository`

### Result
- Close #2923
- Now user can use `findAll(sort, limit)` on `ReactiveSortingRepository`

### DIFF
--- a/src/main/java/org/springframework/data/repository/reactive/ReactiveSortingRepository.java
+++ b/src/main/java/org/springframework/data/repository/reactive/ReactiveSortingRepository.java
@@ -65,7 +65,7 @@ public interface ReactiveSortingRepository<T, ID> extends Repository<T, ID> {
 			throw new IllegalArgumentException("Limit must not be null");
 		}
 
-		if (limit.isUnlimited() || limit.max() == Integer.MAX_VALUE) {
+		if (limit.isUnlimited()) {
 			return findAll(sort);
 		}
 
@@ -76,6 +76,10 @@ public interface ReactiveSortingRepository<T, ID> extends Repository<T, ID> {
 
 		if (maxLimit == 0) {
 			throw new IllegalArgumentException("Limit value cannot be zero");
+		}
+
+		if (maxLimit == Integer.MAX_VALUE) {
+			return findAll(sort);
 		}
 
 		return findAll(sort).take(maxLimit);
--- a/src/main/java/org/springframework/data/repository/reactive/ReactiveSortingRepository.java
+++ b/src/main/java/org/springframework/data/repository/reactive/ReactiveSortingRepository.java
@@ -18,18 +18,19 @@ package org.springframework.data.repository.reactive;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
+import org.springframework.data.domain.Limit;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.repository.NoRepositoryBean;
 import org.springframework.data.repository.Repository;
 
 /**
- * Repository fragment to provide methods to retrieve entities using the sorting abstraction. In many cases it
- * should be combined with {@link ReactiveCrudRepository} or a similar repository interface in order to add CRUD
- * functionality.
+ * Repository fragment to provide methods to retrieve entities using the sorting abstraction. In many cases it should be
+ * combined with {@link ReactiveCrudRepository} or a similar repository interface in order to add CRUD functionality.
  *
  * @author Mark Paluch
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/repository/reactive/ReactiveSortingRepository.java
```diff
@@ -48,4 +49,35 @@ public interface ReactiveSortingRepository<T, ID> extends Repository<T, ID> {
...
+	 * @return all entities sorted by the given options.
+	 * @throws IllegalArgumentException in case the given {@link Sort} is {@literal null} or {@link Limit} is
+	 *           {@literal null} or negative.
+	 */
+	default Flux<T> findAll(Sort sort, Limit limit) {
+		if (limit == null) {
+			throw new IllegalArgumentException("Limit must not be null");
+		}
+
+		if (limit.isUnlimited() || limit.max() == Integer.MAX_VALUE) {
```
injae-kim, line 68:
```suggestion
        final int maxLimit = limit.max(); // Moved from L72
		if (limit.isUnlimited() || maxLimit == Integer.MAX_VALUE) {
```
How about invoke `limit.max` once?

hwan33: When the`max()` method is called on an `Unlimited` instance, it throws an `IllegalStateException`.

```java
@Override
public int max() {
	throw new IllegalStateException(
			"Unlimited does not define 'max'. Please check 'isLimited' before attempting to read 'max'");
}
```

Would it be better if i separate the two conditions into individual if-statements?

injae-kim: oh~ you're right. just ignore my comment 🙇

src/test/java/org/springframework/data/repository/reactive/ReactiveSortingRepositoryTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+			verify(repository).findAll(Sort.unsorted());
+		}
+
+		@DisplayName("Return all entities with maximum integer limit")
+		@Test
+		void shouldReturnAllEntitiesForLimitOfMaxValue() {
+			when(repository.findAll(Sort.unsorted())).thenReturn(Flux.range(1, 20));
+
+			final Flux<Integer> results = repository.findAll(Sort.unsorted(), Limit.unlimited());
```
injae-kim, line 102:
```suggestion
			final Flux<Integer> results = repository.findAll(Sort.unsorted(), Integer.MAX_VALUE);
```
limit should be `MAX_VALUE`, not `unlimited()` here?

hwan33: Oh, Thanks for your good review! I will make the corrections soon.

hwan33:
@mp911de 
Thank you for reviewing my pull request and for the feedback provided. I understand the concerns about aligning with the interface's intent and the potential risks associated with the default implementation using `take` that could lead to loading all data into memory unexpectedly.

Before I move on from this, I would like to clarify a few points to ensure that I fully understand the project's direction and to contribute effectively in the future:

1. Is the main concern that introducing a `limit` parameter does not fit well with the `ReactiveSortingRepository` interface's intended purpose, which is to focus solely on sorting capabilities?

2. If the addition of a `limit` parameter itself is not the issue, is the concern primarily with the default method's implementation that uses `take`? In that case, would an alternative implementation like the following be more aligned with the project's standards?

```java
return findAll(sort)
    .handle((t, sink) -> {
        if (sink.requestedFromDownstream() > 0) {
            sink.next(t);
        }
    })
    .limitRequest(maxLimit);
```

3. If neither the above concern addresses the core issue, could you please provide some guidance or a sample of the preferred implementation approach? It would be greatly beneficial to understand the direction in which we wish to take the repository capabilities, especially concerning limiting the result set.

Your insights will be invaluable for guiding future contributions and ensuring they are in line with the project's goals.

Thank you for your time and consideration.

PR #2963, opened 2023-10-23
Polishing documents

quaff:
Fix https://github.com/spring-projects/spring-data-jpa/issues/3204


mp911de:
Thank you for your contribution. That's merged now.

Issue #2972, opened 2023-11-08
Make AbstractPersistentProperties consider owner for equals.


schauder:
This should resolve https://github.com/spring-projects/spring-data-relational/issues/1657

PR #2973, opened 2023-11-08
AbstractPersistentProperty now considers the owner for equals.

schauder:
This makes a difference when a property is declared in a superclass of two entities.
In such a case the property is the same, but the owner is different.

Closes https://github.com/spring-projects/spring-data-commons/issues/2972
See https://github.com/spring-projects/spring-data-relational/issues/1657

As discussed yesterday I build Spring Data JPA and Spring Data Rest based on this change.
The builds ran without a problem.

I also verify that it actually resolves https://github.com/spring-projects/spring-data-relational/issues/1657

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>3.2.0-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


mp911de (approved):

mp911de:
Please also backport to 3.1.x

odrotbohm:
As per our conversation, LGTM.

schauder:
Merged and backported to 3.1.x

PR #2978, opened 2023-11-16
Resolve LocalDate Storage Inconsistency Issue

hamzahanafi11:
Addressing the inconsistencies in date storage related to MongoDB's default timestamp with timezone approach. The current method often leads to discrepancies in date interpretation across various timezones and services. This commit introduces a revised storage strategy that prioritizes storing dates as LocalDate to ensure consistent interpretation irrespective of timezone differences. By avoiding implicit conversions to timestamps, it mitigates issues where the date representation may differ based on the timezone of the reading service or aggregation functions. This change aims to ensure uniformity in date representation, particularly critical for scenarios like invoicing where the date's day component should be preserved consistently across systems.

### DIFF
--- a/src/main/java/org/springframework/data/convert/Jsr310Converters.java
+++ b/src/main/java/org/springframework/data/convert/Jsr310Converters.java
@@ -45,6 +45,7 @@ import org.springframework.lang.NonNull;
  * @author Christoph Strobl
  * @author Jens Schauder
  * @author Mark Paluch
+ * @author Hamza Hanafi
  */
 public abstract class Jsr310Converters {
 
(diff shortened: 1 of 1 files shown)


mp911de:
These converters are used across Spring Data, not only MongoDB. We intentionally use the system-default zone config as that is something you can control.

If you wish to use store-native JSR-310 support, see https://docs.spring.io/spring-data/mongodb/reference/mongodb/mapping/mapping.html#mapping-configuration and please configure the conversions with `MongoConverterConfigurationAdapter.useNativeDriverJavaTimeCodecs()`.

PR #2985, opened 2023-11-21
Add `@Transient` properties to `PersistentEntity` and use value defaulting for transient constructor properties

### LINKED ISSUE #1432: Include transient properties in persistent entity metamodel [DATACMNS-978]
**[Oliver Drotbohm](https://jira.spring.io/secure/ViewProfile.jspa?name=olivergierke)** opened **[DATACMNS-978](https://jira.spring.io/browse/DATACMNS-978?redirect=false)** and commented

Currently a `PersistentEntity` doesn't even carry around properties that are considered transient. They are already skipped when the entity is populated in the first place. This has caused issues in places where **all** properties needed to be inspected

---
No further details from [DATACMNS-978](https://jira.spring.io/browse/DATACMNS-978?redirect=false)

### LINKED ISSUE #2942: Add ignoring @Transient fields in constructors
This is a feature request. Currently querying an entity like this would fail:

```
@Table(name = "chat__channel")
data class ChatChannel(
    val name: String,
    @Transient val creationRequestId: String? = null
) : BaseEntity()
```

with exception:
```
java.lang.IllegalStateException: Required property creationRequestId not found for class com.gtomato.server.application.messenger.messageservice.entity.ChatChannel
	at org.springframework.data.mapping.PersistentEntity.getRequiredPersistentProperty(PersistentEntity.java:190) ~[spring-data-commons-3.1.2.jar!/:3.1.2]
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException:
Error has been observed at the following site(s):
	*__checkpoint ⇢ Handler com.gtomato.server.application.messenger.messageservice.api.ChannelApi#list(String, Continuation) [DispatcherHandler]
Original Stack Trace:
		at org.springframework.data.mapping.PersistentEntity.getRequiredPersistentProperty(PersistentEntity.java:190) ~[spring-data-commons-3.1.2.jar!/:3.1.2]
		at org.springframework.data.r2dbc.convert.MappingR2dbcConverter$RowParameterValueProvider.getParameterValue(MappingR2dbcConverter.java:715) ~[spring-data-r2dbc-3.1.2.jar!/:3.1.2]
		at org.springframework.data.mapping.model.SpELExpressionParameterValueProvider.getParameterValue(SpELExpressionParameterValueProvider.java:49) ~[spring-data-commons-3.1.2.jar!/:3.1.2]
		at org.springframework.data.relational.core.conversion.BasicRelationalConverter$ConvertingParameterValueProvider.getParameterValue(BasicRelationalConverter.java:298) ~[spring-data-relational-3.1.2.jar!/:3.1.2]
		at org.springframework.data.mapping.model.KotlinClassGeneratingEntityInstantiator$DefaultingKotlinClassInstantiatorAdapter.extractInvocationArguments(KotlinClassGeneratingEntityInstantiator.java:222) ~[spring-data-commons-3.1.2.jar!/:3.1.2]
		at org.springframework.data.mapping.model.KotlinClassGeneratingEntityInstantiator$DefaultingKotlinClassInstantiatorAdapter.createInstance(KotlinClassGeneratingEntityInstantiator.java:196) ~[spring-data-commons-3.1.2.jar!/:3.1.2]
		at org.springframework.data.mapping.model.ClassGeneratingEntityInstantiator.createInstance(ClassGeneratingEntityInstantiator.java:98) ~[spring-data-commons-3.1.2.jar!/:3.1.2]
		at org.springframework.data.relational.core.conversion.BasicRelationalConverter.createInstance(BasicRelationalConverter.java:136) ~[spring-data-relational-3.1.2.jar!/:3.1.2]
		at org.springframework.data.r2dbc.convert.MappingR2dbcConverter.createInstance(MappingR2dbcConverter.java:328) ~[spring-data-r2dbc-3.1.2.jar!/:3.1.2]
		at org.springframework.data.r2dbc.convert.MappingR2dbcConverter.read(MappingR2dbcConverter.java:120) ~[spring-data-r2dbc-3.1.2.jar!/:3.1.2]
		at org.springframework.data.r2dbc.convert.MappingR2dbcConverter.read(MappingR2dbcConverter.java:115) ~[spring-data-r2dbc-3.1.2.jar!/:3.1.2]
		at org.springframework.data.r2dbc.convert.EntityRowMapper.apply(EntityRowMapper.java:42) ~[spring-data-r2dbc-3.1.2.jar!/:3.1.2]
		at org.springframework.data.r2dbc.convert.EntityRowMapper.apply(EntityRowMapper.java:29) ~[spring-data-r2dbc-3.1.2.jar!/:3.1.2]
		at io.asyncer.r2dbc.mysql.MySqlResult.lambda$map$1(MySqlResult.java:94) ~[r2dbc-mysql-1.0.2.jar!/:1.0.2]

...
```

Although there are many workarounds, it would be better if fields marked `@Transient` and have a default value in the constructor could be "ignored", so that we do not have to add another constructor to handle this use case.

mp911de:
We now maintain `@Transient` metadata in the mapping model to determine whether a transient property has been used in a Kotlin Data class, Record or plain Java class. 

```java
record MyRecord(String firstname, @Transient String lastname) {
}
```

```kotlin
data class MyDataClass(@Transient val firstname: String = "foo", @Transient val age: Int)
```

Entity instantiation populates properties with default values (`null` for object types, or primitive default values) so that Kotlin data classes can use defaulting.

Closes #1432 
Closes #2942

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>3.3.0-SNAPSHOT</version>
+	<version>3.3.0-ISSUE-1432-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 	<description>Core Spring concepts underpinning every Spring Data module.</description>
--- a/src/main/antora/modules/ROOT/pages/object-mapping.adoc
+++ b/src/main/antora/modules/ROOT/pages/object-mapping.adoc
@@ -154,7 +154,7 @@ By default, Spring Data attempts to use generated property accessors and falls b
 Let's have a look at the following entity:
 
 .A sample entity
-[source, java]
+[source,java]
 ----
 class Person {
 
@@ -165,14 +165,15 @@ class Person {
 
   private String comment;                                                   <4>
   private @AccessType(Type.PROPERTY) String remarks;                        <5>
+  private @Transient String summary;                                        <6>
 
-  static Person of(String firstname, String lastname, LocalDate birthday) { <6>
+  static Person of(String firstname, String lastname, LocalDate birthday) { <7>
 
     return new Person(null, firstname, lastname, birthday,
       Period.between(birthday, LocalDate.now()).getYears());
   }
 
-  Person(Long id, String firstname, String lastname, LocalDate birthday, int age) { <6>
+  Person(Long id, String firstname, String lastname, LocalDate birthday, int age) { <7>
 
     this.id = id;
     this.firstname = firstname;
@@ -201,7 +202,10 @@ With the design shown, the database value will trump the defaulting as Spring Da
 Even if the intent is that the calculation should be preferred, it's important that this constructor also takes `age` as parameter (to potentially ignore it) as otherwise the property population step will attempt to set the age field and fail due to it being immutable and no `with…` method being present.
 <4> The `comment` property is mutable and is populated by setting its field directly.
 <5> The `remarks` property is mutable and is populated by invoking the setter method.
-<6> The class exposes a factory method and a constructor for object creation.
+<6> The `summary` property transient and will not be persisted as it is annotated with `@Transient`.
+Spring Data doesn't use Java's `transient` keyword to exclude properties from being persisted as `transient` is part of the Java Serialization Framework.
+Note that this property can be used with a persistence constructor, but its value will default to `null` (or the respective primitive initial value if the property type was a primitive one).
+<7> The class exposes a factory method and a constructor for object creation.
 The core idea here is to use factory methods instead of additional constructors to avoid the need for constructor disambiguation through `@PersistenceCreator`.
 Instead, defaulting of properties is handled within the factory method.
 If you want Spring Data to use the factory method for object instantiation, annotate it with `@PersistenceCreator`.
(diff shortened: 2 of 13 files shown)


odrotbohm:
Do we really need `getTransientProperty(…)`? I wonder what the use case is to actually look up those explicitly? Also, dies that mean that `getProperty(…)` only returns non-transient properties now?

One other thing we have to double-check is the expectations that client code using `doWithProperties(…)` had. So far, the callback was not involved for transient properties, simply because they haven't been captured at all. Unfortunately, this means that the code in those callbacks doesn't expect to be called for transient properties in the first place and doesn't guard against that case.

mp911de:
> One other thing we have to double-check is the expectations that client code using doWithProperties(…) had

We never provided transient properties to callbacks via `doWithProperties` and I would not want to change that. Otherwise, a lot of code assumptions get broken.

Likely we don't need to expose `getTransientProperty(…)`. This is merely used internally for the `isTransient` check.

> Also, dies that mean that getProperty(…) only returns non-transient properties now?

`getProperty` never returned transient properties in the first place. It's intended to remain that way. 

```java
	public void treatsTransientAsNotExisting() {
		assertThat(getProperty(ClassWithReadOnlyProperties.class, "transientProperty")).isNull();
	}
```

Issue #2990, opened 2023-11-27
`KotlinBeanInfoFactory` throws an exception when creating bean info for enums


klaasdellschaft:
Hi,

when upgrading my project from Spring Boot 3.1.x to 3.2.0 (and thus from Spring Beans 6.0.x / Spring Data Commons 3.1.x), I get an exception during runtime when the `KotlinBeanInfoFactory` tries to return the bean info for a Kotlin enum type. The following code can be used for reproducing the problem:

```
enum class SomeEnum(val someProperty: String) {
	VALUE_1("someValue1"),
	VALUE_2("someValue2")
}

@Test
fun determinesEnumProperties() {
    val pds = org.springframework.beans.BeanUtils.getPropertyDescriptors(SomeEnum::class.java)
    assertThat(pds).extracting("name").contains("someProperty")
}
```

This produces the following exception:

```
kotlin.reflect.jvm.internal.KotlinReflectionInternalError: No accessor found for property val entries: kotlin.enums.EnumEntries<org.springframework.data.util.KotlinBeanInfoFactoryUnitTests.SomeEnum>

	at kotlin.reflect.jvm.internal.KPropertyImplKt.computeCallerForAccessor(KPropertyImpl.kt:314)
	at kotlin.reflect.jvm.internal.KPropertyImplKt.access$computeCallerForAccessor(KPropertyImpl.kt:1)
	at kotlin.reflect.jvm.internal.KPropertyImpl$Getter$caller$2.invoke(KPropertyImpl.kt:180)
	at kotlin.reflect.jvm.internal.KPropertyImpl$Getter$caller$2.invoke(KPropertyImpl.kt:179)
	at kotlin.SafePublicationLazyImpl.getValue(LazyJVM.kt:107)
	at kotlin.reflect.jvm.internal.KPropertyImpl$Getter.getCaller(KPropertyImpl.kt:179)
	at kotlin.reflect.jvm.ReflectJvmMapping.getJavaMethod(ReflectJvmMapping.kt:64)
	at kotlin.reflect.jvm.ReflectJvmMapping.getJavaGetter(ReflectJvmMapping.kt:49)
	at org.springframework.data.util.KotlinBeanInfoFactory.getBeanInfo(KotlinBeanInfoFactory.java:67)
	at org.springframework.beans.CachedIntrospectionResults.getBeanInfo(CachedIntrospectionResults.java:222)
[...]

PR #2991, opened 2023-11-29
Exclude "entries" property from bean info for kotlin enum class

quaff:
Fix GH-2990


mp911de:
Thanks for the proposed change. We want to back off entirely, therefore 295d14e34d7184b2545d9b7b32398eb976d89c8c contains a slightly different change.

PR #2996, opened 2023-12-04
Introduce `MethodParameterFactory` to `Parameters`

mp911de:
We now provide a factory function to create `MethodParameter` objects associated with the enclosing class so parameters can resolve generics properly.


src/main/java/org/springframework/data/repository/query/DefaultParameters.java
```diff
@@ -40,13 +42,30 @@ public DefaultParameters(Method method) {
...
+	 * resolve generic method parameter types that are declared on the containing class. Use
+	 * {@link DefaultParameters#DefaultParameters(Method, IntFunction, TypeInformation)} with a
+	 * {@link MethodParameter#withContainingClass(Class) contextual factory} to ensure proper gerneric resolution.
 	 *
 	 * @param method must not be {@literal null}.
 	 * @param aggregateType must not be {@literal null}.
 	 */
 	public DefaultParameters(Method method, TypeInformation<?> aggregateType) {
-		super(method, param -> new Parameter(param, aggregateType));
+		this(method, index -> new MethodParameter(method, index), aggregateType);
```
christophstrobl, line 54: should we consider the aggregate type here and pass it on via `withContainingClass`?

src/main/java/org/springframework/data/repository/query/Parameters.java
```diff
@@ -421,7 +444,9 @@ public static boolean isBindable(Class<?> type) {
 		return !TYPES.contains(type);
 	}
 
+	@Override
 	public Iterator<T> iterator() {
 		return parameters.iterator();
 	}
+
```
odrotbohm, line 451: Superfluos?

odrotbohm (commented):
Can you elaborate why you think creating the dependency to the `RepositoryMetadata` is needed? It looks like it's solely used to pull out the domain type on construction? So why not stay with the variant that takes a `TypeInformation` and extract that on creation of the parameters?

mp911de:
We have two dependencies (repository interface, domain type) now that we derive from `RepositoryMetadata`. Instead passing on each one individually with a growing constructor, it makes sense to pass on `RepositoryMetadata`. We already depend on the `repository.core` package so design-wise it seemed a good fit.

odrotbohm:
I might be missing something but neither `Parameters` nor `DefaultParameters` previously depended on some repository abstraction, did they? It's all pure `Method` and `TypeInformation`. So I wonder, why we would now force callers of these constructors to first create a `RepositoryMetadata` that needs a full repository interface, when the `TypeInformation` for the aggreagte type is just enough?

mp911de:
This change isn't about the domain type, it is about `methodParameter.withContainingClass(repositoryMetadata.getRepositoryInterface());`.

mp911de:
Yes, was referring to:

https://github.com/spring-projects/spring-data-commons/pull/2996/files#diff-1378f3880a2db27e418f0a9d7090e5b7868f405555be19a1d6f1491bda15d299R103-R128

PR #2998, opened 2023-12-10
Remove extra `/api` path segement in reference documentation

RajAnthari:
Found a broken link to ListCrudRepository javadoc at https://docs.spring.io/spring-data/relational/reference/repositories/core-concepts.html.
So, I fixed it by removing extra `/api`.

image


mp911de:

PR #3001, opened 2023-12-12
Introduce `@Lookup` methods on repository

quaff:
After this commit, we could lookup beans from BeanFactory and use them in default methods.

```java
interface FooRepository extends CrudRepository<Foo, Long> {

	default void doSomething(Long fooId, Long barId) {
		Optional<Foo> foo = findById(fooId);
		Optional<Bar> bar = barRepository().findById(barId);
		//TODO do something with foo and bar
	}

	@Lookup
	BarRepository barRepository();
}
```

```java
interface BarRepository extends CrudRepository<Bar, Long> {

}
```

### DIFF
--- a/src/main/java/org/springframework/data/repository/core/RepositoryInformation.java
+++ b/src/main/java/org/springframework/data/repository/core/RepositoryInformation.java
@@ -23,6 +23,7 @@ import org.springframework.data.util.Streamable;
  * Additional repository specific information
  *
  * @author Oliver Gierke
+ * @author Yanming Zhou
  */
 public interface RepositoryInformation extends RepositoryMetadata {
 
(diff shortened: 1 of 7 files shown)


mp911de:
That's not how we want our code to be used. You're mixing concerns of a bean container with public API. Use custom repository fragments with `@Autowired` instead.

quaff:
Or we could narrow the type only accept `EntityManager`.
```java
interface FooRepository extends CrudRepository<Foo, Long> {

	default void doSomething(Long fooId) {
		EntityManager entityManager = entityManager();
		//TODO do something with entityManager and fooId
	}

	@Lookup
	EntityManager entityManager();
}
```
It's more concise compare to custom repository fragments, WDYT?

quaff:
@mp911de  Could you guide me the easiest way to implement such extension at application side, customizing `repositoryFactoryBeanClass` of `@EnableJpaRepositories` seems viable but not easy because currently there are not enough extension points, some methods are private and some codes should be extracted to protected method to allow override.

PR #3016, opened 2024-01-04
Polishing

quaff:
1. reuse fields from super class
2. remove unused constant
3. add static modifier to inner class

### DIFF
--- a/src/main/java/org/springframework/data/repository/config/AnnotationRepositoryConfigurationSource.java
+++ b/src/main/java/org/springframework/data/repository/config/AnnotationRepositoryConfigurationSource.java
@@ -51,6 +51,7 @@ import org.springframework.util.StringUtils;
  * @author Mark Paluch
  * @author Johannes Englmeier
  * @author Florian Cramer
+ * @author Yanming Zhou
  */
 public class AnnotationRepositoryConfigurationSource extends RepositoryConfigurationSourceSupport {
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/repository/config/AnnotationRepositoryConfigurationSource.java
```diff
@@ -68,8 +69,6 @@ public class AnnotationRepositoryConfigurationSource extends RepositoryConfigura
 	private final AnnotationMetadata enableAnnotationMetadata;
 	private final AnnotationAttributes attributes;
 	private final ResourceLoader resourceLoader;
-	private final Environment environment;
```
mp911de, line 71: These changes are fine.

src/main/java/org/springframework/data/repository/config/RepositoryConfigurationSourceSupport.java
```diff
@@ -34,14 +34,13 @@
  * @author Thomas Darimont
  * @author Peter Rietzler
  * @author Jens Schauder
+ * @author Yanming Zhou
  */
 public abstract class RepositoryConfigurationSourceSupport implements RepositoryConfigurationSource {
 
-	protected static final String DEFAULT_REPOSITORY_IMPL_POSTFIX = "Impl";
```
mp911de, line 40: These aren't because we do not want to alter visibility.

PR #3054, opened 2024-03-01
Do not register SpringDataWebSettings if pageSerializationMode is default

quaff:
allow application to register their own `SpringDataWebSettings` without `@Primary`

see https://github.com/spring-projects/spring-boot/pull/39797#discussion_r1508396169


odrotbohm:
Thanks for the contribution. This polished and merged.

PR #3058, opened 2024-03-04
Refine Query by Example string matching to call out store-specific limitations

kmeyer-mbs:
Criteria API doesn't support regular expressions.

Related tickets [spring-data-jpa 1295](https://github.com/spring-projects/spring-data-jpa/issues/1295), [spring-data-jpa 3391](https://github.com/spring-projects/spring-data-jpa/issues/3391)


mp911de:
We need to refine our documentation, however, we cannot list each store rules individually as we would end up with a huge matrix providing details for unrelated stores (i.e. the MongoDB documentation or Redis documentation would talk about JPA specifics). I'm going to tweak the wording accordingly.

PR #3072, opened 2024-04-03
Revise `OffsetScrollPosition` to use zero-based indexes

christophstrobl:


src/main/java/org/springframework/data/domain/OffsetScrollPosition.java
```diff
@@ -141,7 +149,7 @@ public OffsetScrollPosition apply(int offset) {
 				throw new IndexOutOfBoundsException(offset);
 			}
 
-			return of(startOffset + offset + 1);
+			return of(startOffset + offset);
```
mp911de, line 152: How about introducing a factory method for `OffsetPositionFunction` on `OffsetScrollPosition` to hide away `offset.isInitial() ? 0 : offset.getOffset()` as in `OffsetScrollPosition.positionFunction(offset.isInitial() ? 0 : offset.getOffset())` on the calling code site?

PR #3075, opened 2024-04-11
Use `-1` instead of `null` as initial offset

quaff:
1. Use type `long` instead of `Long` to keep it consistent with previous version. 
2. Position function is simplified to `offset+1`.


mp911de:

PR #3105, opened 2024-06-10
Fix internal typo of class name

knalli:
Hi there, I have noticed a small typo in the class SpringDataWebSettingsRegistar => SpringDataWebSettingsRegistrar

Because this is a public class, a simple rename could be breaking change. I have added an alias with deprecation hint for removal.

There is no test that covers such a typo. Since I did not change the functionality at all, I did not add myself.


mp911de:
Thanks a lot, that's fixed now with #3178

Issue #3100, opened 2024-05-22
Use Spring's `Nullness` utility to determine JSpecify nullness


Helmsdown:
Greetings. I am upgrading from Spring Boot 2.7 to Spring Boot 3. That means I'm going from `spring-data-commons:2.7.18` to `spring-data-commons:3.2.5`. Additionally this upgrade is bringing along a Hibernate 5.4 to 6.4.4 migration. Hibernate 6+ is all in on `jakarta.annotation` stuff.

In fact the automation my company wrote to help with the upgrade generously refactored all usages in my repo of `javax.annotation` to `jakarta.annotation`.

However, now a JPA query class I have is making `org.springframework.data.repository.core.support.MethodInvocationValidator` sad because the method is annotated with `jakarta.annotation.Nullable` but NullableUtils does evaluate that annotation when deciding whether something is or is not documented as "nullable".

Here are some questions to guide this discussion:
1. It seems that this strategy of "if this one annotation class is on the classpath, then that _must_ be the right one to use" is a bit...myopic. Especially given the code falls back to `org.springframework.lang.Nullable`, `reactor.util.lang.NonNullApi`, and `rg.springframework.lang.NonNullApi` if it can't find `javax.annotation.Nullable`. Is it fair to say "this strategy could be improved upon"?
2. Do you agree that the current state of NullabileUtils is at best a paper cut when upgrading to SB3 and Spring 6 (and at worst a blocker if you use spring-data-jpa)?
3. Would you support an annotation resolution strategy that is as follows?
 a. Evaluate what nullability annotations are on the classpath (javax, jakarta, spring, reactor, spotbugs, findbugs, intellij, etc)
 b. When attempting to evaluate if a given parameter or return type are nullable, consult all annotations that are present on the method or parameter in a deterministic order, if any are present, use them.
 c. bonus points: if there is a disagreement in the annotations (e.g. both Nullable and Nonnull) throw an exception

I look forward to hearing back from you

mp911de:
TL;DR: Supporting JSR-305 annotations (`avax.annotation.Nullable`) instead of the Jakarta packages is a consequence of Spring Framework's support for these annotations. It is not only the meta-annotation `org.springframework.lang.Nullable` but also runtime support for annotation discovery.

[...]

PR #3115, opened 2024-06-26
Introduce `Nullability` API

mp911de:
We now provide a dedicated API to introspect nullability rules. The new API supports Spring, JSR-305, and [JSpecify](https://jspecify.dev/).

```java
Nullability nullability = Nullability.forMethodReturnType(method);
nullability.isDeclared();
nullability.isNullable();
nullability.isNonNull();

// introspect multiple elements for their nullability in the scope of a class/package.
Nullability.Introspector introspector = Nullability.introspect(NonNullOnPackage.class);
Nullability nullability = introspector.forReturnType(method);
```

The previous `NullableUtils` API is now deprecated as its implementation would not support a pluggable model of `NullabilityProvider`.

Related to #3100


mp911de:
Paging @sdeleuze

odrotbohm:
Should this rather live in Spring Framework?

sdeleuze:
If we added more nullabilty runtime API, it would indeed be worth to considering it for inclusion in Spring Framework, but in terms of timeframe it would be more towards next year. As shared with @mp911de, nullabilty is a complex and nuanced topic with a lot of implications, and there are some ongoing discussions with JSpecify working group on potential evolution that could impact how Spring supports it, as well as some future implication on JDK side. With @jhoeller, in terms of timeline we suggest to wait for related portfolio-wide discussions to evaluate what could go on Framework side before merging this PR.

mp911de:
Spring Framework introduced a `Nullness` utility (see also #3100) so we can leverage that one to detect null markers instead of having an own utility.

PR #3142, opened 2024-08-25
Changed the access level of AbstractAggregateRoot's methods to public

mipo256:
This PR consists of two things:

1. Changing the access level of API methods to public in `AbstractAggregateRoot`. The problem is that this class is supposed to be extended, but because the API access level is `protected` [the following is required](https://github.com/mipo256/spring-data-features-we-do-not-know-about/blob/master/src/main/java/io/mpolivaha/domain_events/CargoEntity.java#L40). That is a bit awkward to override the method to just call the `super` implementation. 

2. I've created a single instance of the `IllegalStateException` exception to throw in case of domain events change. This is to optimize the stacktrace creation and thus the performance of this case. Since the exception was pure for control flow, it does not break the contract or change the behavior. [Reference](https://shipilev.net/blog/2014/exceptional-performance/#Rationale).

### DIFF
--- a/src/main/java/org/springframework/data/domain/AbstractAggregateRoot.java
+++ b/src/main/java/org/springframework/data/domain/AbstractAggregateRoot.java
@@ -31,6 +31,7 @@ import org.springframework.util.Assert;
  *
  * @author Oliver Gierke
  * @author Christoph Strobl
+ * @author Mikhail Polivakha
  * @since 1.13
  */
 public class AbstractAggregateRoot<A extends AbstractAggregateRoot<A>> {
@@ -44,7 +45,7 @@ public class AbstractAggregateRoot<A extends AbstractAggregateRoot<A>> {
 	 * @return the event that has been added.
 	 * @see #andEvent(Object)
 	 */
-	protected <T> T registerEvent(T event) {
+	public <T> T registerEvent(T event) {
 
 		Assert.notNull(event, "Domain event must not be null");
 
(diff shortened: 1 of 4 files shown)


odrotbohm:
Thanks for submitting this pull request. My first impression is that we're not going to go forward with this, and here are the reasons why:

1. `AbstractAggregateRoot` is designed to be extended and events are supposed to be registered _while issuing state transitions on the aggregate instance. The code you show [here](https://github.com/mipo256/spring-data-features-we-do-not-know-about/blob/49e12bc2de1efc966eb4efd3587145d0f79dceec/src/main/java/io/mpolivaha/domain_events/CargoService.java#L21-L29) resembles a procedural approach: you treat the aggregate like a data container that you get and set data from. That's not how aggregates work. A more idiomatic approach would simply be to add the event in the aggregate method that performs the state change. An example of this can be found [here](https://github.com/odrotbohm/arch-evident-spring/blob/50f6de84ca698ef2e3316d9c932e89a2f59ee482/src/main/java/example/order/Order.java#L57-L64). In that sense, the methods being declared as `protected` is expression an intention of their usage, and not a mistake.
2. I don't think this a spot that needs optimizing. The case in which the exception is created is a programming mistake and should rarely occur. If this case occurs in a way that the exception is triggered at a high rate, the problem is the application code, not the creation of the exception. Furthermore, creating the exception in the constructor captures a completely different stack trace, which is likely to confuse users as it captures the application bootstrap call stack, not the one of the actual invocation.

mipo256:
Thanks for clarification, that makes sense @odrotbohm :)

Issue #3155, opened 2024-09-17
`PageableHandlerMethodArgumentResolver` should resolve `Optional<Pageable>` arguments


vhbcm:
I have OpenAPI generated controllers that use `Optional` for non-required parameters. This results in API like this where, e.g., `Pageable` is wrapped in the `Optional`:

```java
    @RequestMapping(
            method = RequestMethod.GET,
            value = "/discovery/agent",
            produces = {"application/json"})
    default ResponseEntity<DiscoveryAgentPageRestDto> _discoveryAgentGet(
            @Parameter(name = "pageable", description = "Defines page number, page size, sort property and order", schema = @Schema(description = "")) @Valid Optional<Pageable> pageable,
            @Parameter(name = "q", description = "Search query", schema = @Schema(description = "")) @Valid @RequestParam(value = "q", required = false) Optional<String> q)
            throws Exception {
        return discoveryAgentGet(pageable, q);
    }
```

However, the stock Spring Framework `PageableHandlerMethodArgumentResolver` does not handle this. `PageableHandlerMethodArgumentResolver` should resolve `Optional<Pageable>` arguments

Cross reference: https://github.com/spring-projects/spring-framework/issues/17589#issuecomment-2017508172

quaff:
It's too risky to change method signature of `PageableHandlerMethodArgumentResolver`, I created dedicated resolver for `Optional<Pageable>`, see #3156.

mp911de:
We do not have a case of an absent `Pageable`, even an unpaged `Pageable` can contain some sort. Spring Data represents an empty pageable through `Pageable.unpaged()` and we do not want to introduce another alternative as that makes its usage much more difficult and introduces ambiguity over how the API should be used.

Instead, the OpenAPI part should be fixed.

PR #3156, opened 2024-09-19
Add support for resolving `Optional<Pageable>`

quaff:
Fix GH-3155


mp911de:
We do not have a case of an absent `Pageable`, even an unpaged `Pageable` can contain some sort. Spring Data represents an empty pageable through `Pageable.unpaged()` and we do not want to introduce another alternative as that makes its usage much more difficult and introduces ambiguity over how the API should be used.

PR #3160, opened 2024-09-21
Compute `ReturnType#isDto` eagerly

mipo256:
In order to avoid recalculation of `isDto` for `ReturnedType` it makes sense to wrap the computed result with `Lazy`, giving the fact that `type` is actually final.


mp911de:
`isDto` is computed only once and doesn't change so we can safely pre-compute the value without the need for `Lazy`. I'm going to apply that change during the merge.

mp911de:

PR #3162, opened 2024-09-23
Polishing

ngocnhan-tran1996:
This PR includes

- Use pattern matching
- Polish diamond operator usage
- Consistently isEmpty usage


mp911de:
Thank you for your contribution. That's merged, polished, and backported now. Trivial changes are not attributed with `@author` tags so we rolled these back.

PR #3174, opened 2024-10-09
Update documentation. 

christophstrobl:
Add section on reserved method names within repository interfaces.

Related to: #3173

### DIFF
--- a/src/main/antora/modules/ROOT/pages/repositories/core-concepts.adoc
+++ b/src/main/antora/modules/ROOT/pages/repositories/core-concepts.adoc
@@ -4,6 +4,18 @@
 The central interface in the Spring Data repository abstraction is `Repository`.
 It takes the domain class to manage as well as the identifier type of the domain class as type arguments.
 This interface acts primarily as a marker interface to capture the types to work with and to help you to discover interfaces that extend this one.
+
+[TIP]
+====
+Spring Data considers domain types to be entities, more specifically aggregates.
+So you will see the term "entity" used throughout the documentation that can be interchanged with the term "domain type" or "aggregate".
+
+As you might have noticed in the introduction it already hinted towards domain-driven concepts.
+We consider domain objects in the sense of DDD.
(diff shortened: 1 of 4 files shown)


src/main/antora/modules/ROOT/pages/repositories/core-concepts.adoc
```diff
@@ -37,6 +37,11 @@ public interface CrudRepository<T, ID> extends Repository<T, ID> {
 The methods declared in this interface are commonly referred to as CRUD methods.
 `ListCrudRepository` offers equivalent methods, but they return `List` where the `CrudRepository` methods return an `Iterable`.
 
+[IMPORTANT]
+====
+The repository interface implies a few reserved methods like `findById(ID identifier)` that always target the domain types identifier property regardless of its property name. Read more about this in "`xref:repositories/query-methods-details.adoc#repositories.query-methods.reserved-methods[Defining Query Methods]`".
```
mp911de, line 42: We need to watch for broken references in case a Data module doesn't ship a `repositories/query-methods-details.adoc` documentation page.

christophstrobl: is there one you're aware of right off the top of your head?

src/main/antora/modules/ROOT/pages/repositories/core-concepts.adoc
```diff
@@ -4,6 +4,18 @@
 The central interface in the Spring Data repository abstraction is `Repository`.
 It takes the domain class to manage as well as the identifier type of the domain class as type arguments.
 This interface acts primarily as a marker interface to capture the types to work with and to help you to discover interfaces that extend this one.
+
+[TIP]
+====
+Spring Data considers domain types to be entities, more specifically aggregates.
+So you will see the term "entity" used throughout the documentation that can be interchanged with the term "domain type" or "aggregate".
```
mp911de, line 11: @odrotbohm any opinion on wording? We want to make the point that identifiers need sometimes a different handling than properties, yet in both cases we need to express access to these via some name.

PR #3176, opened 2024-10-14
Move `RepositoryMethodContext` to `repository.core` package

odrotbohm:
Details in the commit message.


src/main/java/org/springframework/data/repository/config/RepositoryConfigurationExtensionSupport.java
```diff
@@ -106,7 +109,12 @@ public String getDefaultNamedQueryLocation() {
 	}
 
 	public void registerBeansForRoot(BeanDefinitionRegistry registry,
-			RepositoryConfigurationSource configurationSource) {}
+			RepositoryConfigurationSource configurationSource) {
+
+		// A proxy RepositoryMethodContext for dependency injection
+		registry.registerBeanDefinition("repositoryMethodContextFactory",
```
mp911de, line 115: Likely we need to check whether there is already such a bean in case an application uses multiple Spring Data modules or a `@Enable…Repositories` is used twice.

odrotbohm: Updated.

src/main/java/org/springframework/data/repository/config/RepositoryConfigurationExtensionSupport.java
```diff
@@ -106,7 +109,13 @@ public String getDefaultNamedQueryLocation() {
 	}
 
 	public void registerBeansForRoot(BeanDefinitionRegistry registry,
-			RepositoryConfigurationSource configurationSource) {}
+			RepositoryConfigurationSource configurationSource) {
+
+		// A proxy RepositoryMethodContext for dependency injection
+		registerIfNotAlreadyRegistered(
```
christophstrobl, line 115: Instance supplier is not supported in AOT and will raise `org.springframework.beans.factory.aot.AotBeanProcessingException`.

src/main/java/org/springframework/data/repository/core/support/DefaultRepositoryMethodContext.java
```diff
@@ -46,15 +50,41 @@ class DefaultRepositoryMethodContext implements RepositoryMethodContext {
...
+		return new DefaultRepositoryMethodContext(AbstractRepositoryMetadata.getMetadata(method.getDeclaringClass()),
+				method);
+	}
+
+	/**
+	 * Creates a proxy {@link RepositoryMethodContext} instance suitable for dependency injection.
+	 *
+	 * @return will never be {@literal null}.
+	 */
+	public static RepositoryMethodContext getInjectionProxy() {
```
christophstrobl, line 72: we'll need additional native image proxy configuration for `RepositoryMethodContext`.

mp911de:

PR #3186, opened 2024-10-24
Explore performance optimizations

### LINKED ISSUE #3185: Explore performance optimizations
We create quite some objects and use heavy concurrent code in code paths that are primarily for read-only operations. We should explore whether we can optimize for less GC pressure.

mp911de:
Closes #3185

### DIFF
--- a/src/main/java/org/springframework/data/util/ClassTypeInformation.java
+++ b/src/main/java/org/springframework/data/util/ClassTypeInformation.java
@@ -31,14 +31,18 @@ import org.springframework.util.ConcurrentLruCache;
  *
  * @author Oliver Gierke
  * @author Christoph Strobl
+ * @author Mark Paluch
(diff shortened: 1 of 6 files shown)


src/main/java/org/springframework/data/util/ClassTypeInformation.java
```diff
@@ -70,7 +74,7 @@ public class ClassTypeInformation<S> extends TypeDiscoverer<S> {
 	 */
 	@Deprecated
 	public static <S> ClassTypeInformation<S> from(Class<S> type) {
-		return from(ResolvableType.forClass(type));
+		return from(resolvableTypeCache.get(type));
```
christophstrobl, line 77: 👍 - I just realised the `ResolvableType` internal cache does not cover this part.

PR #3187, opened 2024-10-24
Fix NPE in `CrudRepositoryExtensions`

OsaSoft:
Erasure of nullability information was causing NPEs in the forced non-null assertion, as it allowed to pass a nullable value, see following snippet that would pass compilation correctly, but throw an unexpected NPE:

```
interface Repo : ListCrudRepository<User, String>

	val repo: Repo = mockk()

	@Test
	fun `CrudRepository#findByIdOrNull() extension should not throw NPE`() {
		val id: String? = null
		// compiler allows null value to be passed to findByIdOrNull, so NPE is thrown
		// I believe this is we lose the non-null type info when the repo extends a java interface,
		// which extends CrudRepository
		repo.findByIdOrNull(id)
	}
```

The fix enforces the generic `ID` type to be a not-null type, therefore the above code no longer compiles. 
image

Not sure how to add a test for this, since the fix is on a static-type level and the problematic code no longer compiles. Could add in a comment, I guess?

### DIFF
--- a/src/main/kotlin/org/springframework/data/repository/CrudRepositoryExtensions.kt
+++ b/src/main/kotlin/org/springframework/data/repository/CrudRepositoryExtensions.kt
@@ -21,6 +21,7 @@ package org.springframework.data.repository
  * @param id the entity id.
  * @return the entity with the given id or `null` if none found
  * @author Sebastien Deleuze
+ * @author Oscar Hernandez
  * @since 2.1.4
  */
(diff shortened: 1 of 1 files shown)


mp911de:
The type should be non-nullable indeed. We would incorporate this change for the next GA release only without backporting it as this is a breaking change in some sense regarding compiling.

efenderbosch-atg:
Just got bit by this today. Any idea when this might get released?

mp911de:
Thank you for your contribution. That's merged now.

PR #3188, opened 2024-10-24
Remove outdated information from Javadoc for PageableHandlerMethodArgumentResolver[Support]

sryze:
The information about default parameter prefix seems outdated. It looks like that portion of the comment was added back in the days when the paging parameters were named `page.page` and `page.size` by default and now they are just `page` and `size` by default. Also there is a missing whitespace between identifiers in HTML docs:
Screenshot 2024-10-25 at 00 09 54


sryze:
Is there a way to sign the CLA without giving access to all account email addresses? I think it actually needs only one email.

mp911de:
CLA isn't required as this is a trivial fix.

Issue #3212, opened 2024-11-25
SPI fragment can not be introduced by `@NoRepositoryBean`


quaff:
Thanks to https://github.com/spring-projects/spring-data-commons/commit/4df7a164e9040bb94dddd5133802db5494856d5f, Spring Data can introduce SPI fragments by specifying them in `spring.factories`, it works fine if repository extends the SPI interface, but it will fails if repository extends an `@NoRepositoryBean` base repository which extends the SPI interface.

Given:
```java
public interface Fragment {
	String getName();
}

public class FragmentImpl implements Fragment {
	public String getName() {
		return "fragment";
	}
}
```
---
Following interface hierarchies works:
```java
@NoRepositoryBean
interface ExcludedRepository extends JpaRepository<TestEntity, Long> {
}

interface TestEntityRepository extends ExcludedRepository, Fragment {
}
```
---
Following interface hierarchies fails:
```java
@NoRepositoryBean
interface ExcludedRepository extends JpaRepository<TestEntity, Long>, Fragment {
}

interface TestEntityRepository extends ExcludedRepository {
}
```

```
Caused by: org.springframework.data.mapping.PropertyReferenceException: No property 'getName' found for type 'TestEntity'
	at org.springframework.data.mapping.PropertyPath.<init>(PropertyPath.java:94) ~[spring-data-commons-3.4.0.jar:3.4.0]
	at org.springframework.data.mapping.PropertyPath.create(PropertyPath.java:455) ~[spring-data-commons-3.4.0.jar:3.4.0]
	at org.springframework.data.mapping.PropertyPath.create(PropertyPath.java:431) ~[spring-data-commons-3.4.0.jar:3.4.0]
	at org.springframework.data.mapping.PropertyPath.lambda$from$0(PropertyPath.java:384) ~[spring-data-commons-3.4.0.jar:3.4.0]
	at java.base/java.util.concurrent.ConcurrentMap.computeIfAbsent(ConcurrentMap.java:330) ~[na:na]
	at org.springframework.data.mapping.PropertyPath.from(PropertyPath.java:366) ~[spring-data-commons-3.4.0.jar:3.4.0]
	at org.springframework.data.mapping.PropertyPath.from(PropertyPath.java:344) ~[spring-data-commons-3.4.0.jar:3.4.0]
	at org.springframework.data.repository.query.parser.Part.<init>(Part.java:81) ~[spring-data-commons-3.4.0.jar:3.4.0]
[...]

PR #3213, opened 2024-11-25
Register indirect fragment introduced by super interfaces

quaff:
Fix GH-3212


mp911de:
We do not want to introduce additional lookups from super-interfaces.

PR #3218, opened 2024-12-09
Removed deprecated api usages

mipo256:


src/main/java/org/springframework/data/repository/query/QueryMethod.java
```diff
@@ -86,7 +86,7 @@ public QueryMethod(Method method, RepositoryMetadata metadata, ProjectionFactory
 		this.method = method;
 		this.unwrappedReturnType = potentiallyUnwrapReturnTypeFor(metadata, method);
 		this.metadata = metadata;
-		this.parameters = createParameters(method, metadata.getDomainTypeInformation());
```
mp911de, line 89: This one needs to stay as-is to ease compatibility. We cannot assume that all stores have implemented `createParameters(ParametersSource parametersSource)`.

PR #3226, opened 2025-01-08
Refining QBE documentation

schauder:


src/main/antora/modules/ROOT/pages/query-by-example.adoc
```diff
@@ -36,6 +36,7 @@ Query by Example is well suited for several use cases:
 Query by Example also has several limitations:
 
 * No support for nested or grouped property constraints, such as `firstname = ?0 or (firstname = ?1 and lastname = ?2)`.
+* No support for matching collections or maps.
```
mp911de, line 39: This limitation only applies to JPA and JDBC. MongoDB does not have this limitation. We could make the bullet point conditional to switch it on/off as needed.

schauder: Ok, will do so.

PR #3231, opened 2025-01-08
Use `List.of()` and `Set.of()` where feasible

quaff:


kse-music:
Note that Arrays.asList is mutable and List.ofis not. Switching to an immutable Collection would be considered as a breaking change.

quaff:
> Note that Arrays.asList is mutable and List.of is not. Switching to an immutable Collection would be considered as a breaking change.

So I didn't replace all `Arrays.asList()`.

mp911de:

PR #3241, opened 2025-02-09
Introduce `KIterablePropertyPath`

### LINKED ISSUE #3010: `KPropertyPath` doesn't support `Iterables`
The `KPropertyPath` currently doesn't support references from `Iterables` such as `List`, `Set`, etc. to the type of the element.

```kotlin
data class User(
    val name: String,
    val addresses: List<Address>
) {
    data class Address(
        val street: String
    )
}

val path = User::addresses / User.Address::street // a type mismatch between the expected the `Address` and offered `List<Address>` occurs here
```

The implementation in `org.springframework.data.mapping.KPropertyPath` should be able to construct paths such as `addresses.street` for the usage in indexes, queries and aggregations to leverage the type-safety of Kotlin for paths references in Spring Data MongoDB.

A similar implementation is offered by the now deprecated project [KMongo](https://github.com/Litote/kmongo/tree/master), see [KPropertyPath.kt](https://github.com/Litote/kmongo/blob/master/kmongo-property/src/main/kotlin/org/litote/kmongo/property/KPropertyPath.kt) and [Properties.kt](https://github.com/Litote/kmongo/blob/master/kmongo-property/src/main/kotlin/org/litote/kmongo/Properties.kt).

mipo256:
Fixes #3010. Fixes the issue in the type safe way, so that this now works :

```
class User(val addresses: List<Address>)
class Address(val street: String)

User::addresses / Author::street contains "Austin" 
```

Where `toDotPath()` renders to `addresses.street`. In order to solve this ticket in the type safe way, without easing the generic boundary from `KProperty1<T, U>` to `KProperty1<Any, U>`, a new `KIterablePropertyPath` needs to be introduced. The solution with `KProperty1<Any, U>` allows for non type safe usages of the API, which is not generally a good idea to begin with.

### DIFF
--- a/src/main/kotlin/org/springframework/data/mapping/KPropertyPath.kt
+++ b/src/main/kotlin/org/springframework/data/mapping/KPropertyPath.kt
@@ -32,13 +32,28 @@ private class KPropertyPath<T, U>(
 ) : KProperty<T> by child
 
 /**
+ * Abstraction of a property path that consists of parent [KProperty],
+ * and child property [KProperty], where parent [parent] has an [Iterable]
+ * of children, so it represents 1-M mapping, not 1-1, like [KPropertyPath]
+ *
+ * @author Mikhail Polivakha
+ */
+internal class KIterablePropertyPath<T, U>(
+	val parent: KProperty<Iterable<U?>?>,
+	val child: KProperty1<U, T>
+) : KProperty<T> by child
+
+/**
  * Recursively construct field name for a nested property.
  * @author Tjeu Kayim
+ * @author Mikhail Polivakha
  */
 internal fun asString(property: KProperty<*>): String {
 	return when (property) {
 		is KPropertyPath<*, *> ->
 			"${asString(property.parent)}.${property.child.name}"
+        is KIterablePropertyPath<*, *> ->
+            "${asString(property.parent)}.${property.child.name}"
 		else -> property.name
 	}
 }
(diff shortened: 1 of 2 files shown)


src/test/kotlin/org/springframework/data/mapping/KPropertyPathTests.kt
```diff
@@ -106,4 +115,8 @@ class KPropertyPathTests {
 
 	class Book(val title: String, val author: Author)
 	class Author(val name: String)
+
+    class User(val addresses: List<Address>)
+
+    class Address(val street: String)
```
mp911de, line 121: Nit: Space indents

src/main/kotlin/org/springframework/data/mapping/KPropertyPath.kt
```diff
@@ -31,14 +31,29 @@ private class KPropertyPath<T, U>(
...
 /**
  * Recursively construct field name for a nested property.
  * @author Tjeu Kayim
+ * @author Mikhail Polivakha
  */
 internal fun asString(property: KProperty<*>): String {
 	return when (property) {
 		is KPropertyPath<*, *> ->
 			"${asString(property.parent)}.${property.child.name}"
+        is KIterablePropertyPath<*, *> ->
```
mp911de, line 55: Nit: Space indents

mp911de:

Issue #3242, opened 2025-02-11
Kotlin Interface projections nullabillity issues


iPave:
Hi, I'm writing an interface projection and instead of getting some readable error while initializing the result I'm getting a runtime NullPointerException on non-nullable type.

```
@Entity
class Customer(
    @Id
    val id: Long? = null,

    var name: String? = null,

    val age: Int
)

interface CustomerRepository : CrudRepository<Customer, Int> {


    @Query("select c.name from Customer c")
    fun customProjectionQuery(): List<NameWithAgeOnly>

}

interface NameWithAgeOnly {
    val name: String
    val age: Int
}

@Service
class Service(
    private val customerRepository: CustomerRepository
) {

    @PostConstruct
    fun test() {
        val customerWithName = Customer(id = 1, name = "John", age = 25)
        val customerWithoutName = Customer(id = 2, name = null, age = 33)
        customerRepository.save(customerWithName)
        customerRepository.save(customerWithoutName)
        val customers = customerRepository.customProjectionQuery()
        for (customer in customers) {
            println(customer.name)
        }
    }
}
```

Caused by: java.lang.NullPointerException: Cannot invoke "org.dev.NameWithAgeOnly.getName()" because "customer" is null

Can you please clarify what am I doing wrong. And maybe give some workaround. Isn't it a bug?
I've also tried to use kotlin dto class instead of interface, but then I'm getting

Caused by: org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type [org.springframework.data.jpa.repository.query.AbstractJpaQuery$TupleConverter$TupleBackedMap] to type [org.dev.NameWithAgeOnly]
	
Here is reproducible example: https://github.com/iPave/jpa-projections-issue/tree/main

quaff:
It's caused by missing column aliases from tuple metadata, not sure Spring Data JPA could handle it, you can fix it by adding alias in your query like `c.name as name`.

IIRC, `data class` will be supported since 4.0.

BTW, you forgot `c.age` in your query.

iPave:
> It's caused by missing column aliases from tuple metadata, not sure Spring Data JPA could handle it, you can fix it by adding alias in your query like `c.name as name`.
> 
> IIRC, `data class` will be supported since 4.0.
> 
> BTW, you forgot `c.age` in your query.

Thanks for you reply, even if I add aliases jpa just puts null to notNullable property and this code 

```    @PostConstruct
    fun test() {
[...]

PR #3244, opened 2025-02-14
Throw exception if Kotlin projection requires non-null value but null result present

quaff:
Fix GH-3242

### DIFF
--- a/src/main/java/org/springframework/data/projection/ProjectingMethodInterceptor.java
+++ b/src/main/java/org/springframework/data/projection/ProjectingMethodInterceptor.java
@@ -16,6 +16,7 @@
 package org.springframework.data.projection;
 
 import java.lang.reflect.Array;
+import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/projection/ProjectingMethodInterceptor.java
```diff
@@ -83,6 +90,14 @@ public Object invoke(@SuppressWarnings("null") @NonNull MethodInvocation invocat
 			return conversionService.convert(new NullableWrapper(result), typeToReturn.getType());
 		}
 
+		if (result == null) {
+			KFunction<?> function = KotlinDetector.isKotlinType(method.getDeclaringClass()) ?
+					KotlinReflectionUtils.findKotlinFunction(method) : null;
+			if (function != null && !function.getReturnType().isMarkedNullable()) {
+				throw new IllegalArgumentException("Kotlin function '%s' requires non-null return value".formatted(method.toString()));
```
quaff, line 97: Not sure exception type and message is appropriate.

christophstrobl:
Thank you @quaff! Merged to _main_ development line.

PR #3259, opened 2025-03-21
Expose query method metadata

mipo256:
I think it is worth to make the accessors `public` (or `protected` maybe) for method and `RepositoryMetadata`. The Spring Data JDBC for instance, because it cannot access the `method` field in the `QueryMethod` [has to re-declare it inside the `JdbcQueryMethod`](https://github.com/spring-projects/spring-data-relational/blob/main/spring-data-jdbc/src/main/java/org/springframework/data/jdbc/repository/query/JdbcQueryMethod.java#L58), so it is clearly redundant, since the method already is present in the parent.

P.S: This is also would help us implemented the `YdbQueryMethod` in the [Spring Data JDBC dialect for YDB database](https://github.com/ydb-platform/ydb-java-dialects/issues/173). 

CC: @mp911de

### DIFF
--- a/src/main/java/org/springframework/data/mapping/model/PersistentEntityIsNewStrategy.java
+++ b/src/main/java/org/springframework/data/mapping/model/PersistentEntityIsNewStrategy.java
@@ -41,18 +41,21 @@ class PersistentEntityIsNewStrategy implements IsNewStrategy {
 	 * Creates a new {@link PersistentEntityIsNewStrategy} for the given entity.
 	 *
 	 * @param entity must not be {@literal null}.
+	 * @param idOnly {@code true} if should consider only the ID property of the {@link PersistentEntity}, {@code false}
+	 *               if other properties such as {@link PersistentEntity#getVersionProperty() version} shall be
+	 *               considered.
 	 */
 	private PersistentEntityIsNewStrategy(PersistentEntity<?, ?> entity, boolean idOnly) {
 
 		Assert.notNull(entity, "PersistentEntity must not be null");
 
-		this.valueLookup = entity.hasVersionProperty() && !idOnly //
-				? source -> entity.getPropertyAccessor(source).getProperty(entity.getRequiredVersionProperty())
-				: source -> entity.getIdentifierAccessor(source).getIdentifier();
-
-		this.valueType = entity.hasVersionProperty() && !idOnly //
-				? entity.getRequiredVersionProperty().getType() //
-				: entity.hasIdProperty() ? entity.getRequiredIdProperty().getType() : null;
+		if (entity.hasVersionProperty() && !idOnly) {
+			this.valueLookup = source -> entity.getPropertyAccessor(source).getProperty(entity.getRequiredVersionProperty());
+			this.valueType = entity.getRequiredVersionProperty().getType();
+		} else {
+			this.valueLookup = source -> entity.getIdentifierAccessor(source).getIdentifier();
+			this.valueType = entity.hasIdProperty() ? entity.getRequiredIdProperty().getType() : null;
+		}
 
 		Class<?> type = valueType;
 
@@ -60,8 +63,9 @@ class PersistentEntityIsNewStrategy implements IsNewStrategy {
 
 			if (!ClassUtils.isAssignable(Number.class, type)) {
 
-				throw new IllegalArgumentException(String
-						.format("Only numeric primitives are supported as identifier / version field types; Got: %s", valueType));
+				throw new IllegalArgumentException(
+						String.format("Only numeric primitives are supported as identifier / version field types; Got: %s",
+								valueType));
 			}
 		}
 	}
--- a/src/main/java/org/springframework/data/repository/query/Parameters.java
+++ b/src/main/java/org/springframework/data/repository/query/Parameters.java
@@ -38,6 +38,8 @@ import org.springframework.util.Assert;
 /**
  * Abstracts method parameters that have to be bound to query parameters or applied to the query independently.
  *
+ * @param <T> The exact type of {@link Parameter} inside {@link Parameters}
+ * @param <S> The exact type of {@link Parameters} to be used. Generified to allow for recursive generics.
  * @author Oliver Gierke
  * @author Christoph Strobl
  * @author Johannes Englmeier
--- a/src/main/java/org/springframework/data/repository/query/QueryMethod.java
+++ b/src/main/java/org/springframework/data/repository/query/QueryMethod.java
@@ -86,7 +86,7 @@ public class QueryMethod {
 		this.method = method;
 		this.unwrappedReturnType = potentiallyUnwrapReturnTypeFor(metadata, method);
 		this.metadata = metadata;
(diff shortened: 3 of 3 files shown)


mp911de:
> QueryMethod [has to re-declare it inside the JdbcQueryMethod](https://github.com/spring-projects/spring-data-relational/blob/main/spring-data-jdbc/src/main/java/org/springframework/data/jdbc/repository/query/JdbcQueryMethod.java#L58),

This is by design. We do not want to expose constructor arguments as `public` API. If you want to expose these, please define appropriate methods in your implementation class.

However, `JdbcQueryLookupStrategy` (creating `JdbcQueryMethod`) ships with a multitude of context (8 constructor args) and it is package-private. How do you want to create your own `YdbQueryMethod`? I don't really see how that would work.

Going forward, we do not provide SQL types or type names for `MapSqlParameterSource` when binding parameters (specifically expressions or scalar values). The only thing that comes to my mind is using `JdbcValue`. You could add a YdbLimit class to your dialect (`YdbLimit.of(…)`).

mipo256:
> This is by design. We do not want to expose constructor arguments as public API. If you want to expose these, please define appropriate methods in your implementation class.

Okay, I get it that this is an _arguable_ idea to expose the `method` and `repositoryMetadata` via public getters, I kind of agree. 

But let's at least make the getters `protected`. The child of course aware about the `method` and `repositoryMetadata`, since it provided them to the parent via `super()` call, but in this case the child has to re-declare the `method` and `repositoryMetadata` inside of it. 

So we basically needlessly waste space for two more references, without getting any extra benefits. And this complicates the implementation of the child, btw.

mipo256:
> However, JdbcQueryLookupStrategy (creating JdbcQueryMethod) ships with a multitude of context (8 constructor args) and it is package-private. How do you want to create your own YdbQueryMethod? I don't really see how that would work.

Well, we do not use the `JdbcQueryLookupStrategy`, we extend it, [see here the corresponding class in the PR  ](https://github.com/ydb-platform/ydb-java-dialects/pull/181/files#diff-d7f3eb1fd5295c12d964853abb1bfff142af528d4ac43039b105d55ac556b8d4R91).

And, again, we of course can re-declare the references to fields that we pass to parent via `super()` - but why? I do not really see any compelling reason to do so to be honest. That is why [the following change was proposed](https://github.com/spring-projects/spring-data-relational/pull/2000/files#diff-d7f330015c68b820d348d2cbe19dd5d3795b63cb673304317d721ab022118b42R273). I think it is fine to give that _to children_, they need it.

mipo256:
> Going forward, we do not provide SQL types or type names for MapSqlParameterSource when binding parameters (specifically expressions or scalar values).

I am aware of that. That is why we have overridden the `createParamaters`. 

>  The only thing that comes to my mind is using JdbcValue. You could add a YdbLimit class to your dialect (YdbLimit.of(…)).

Unfortunately, that would not help. Changes in dialect are static. The idea is that the user must be able to dynamically provide the `SQLType` of the underlying parameter.

mp911de:
> But let's at least make the getters protected. The child of course aware about the method and repositoryMetadata, since it provided them to the parent via super() call, but in this case the child has to re-declare the method and repositoryMetadata inside of it.

If a subclass wants to expose these, then it should introduce dedicated methods.

> And, again, we of course can re-declare the references to fields that we pass to parent via super() - but why? I do not really see any compelling reason to do so to be honest. 

Encapsulation and design. Not everything a constructor receives is intended to be exposed publicly. We hide certain aspects to avoid coupling (everything that is even remotely visible is going to be used. Everything being used and then changed generates tickets that ask for a reply).

> Well, we do not use the JdbcQueryLookupStrategy directly, we extend it, [see here the corresponding class in the PR ](https://github.com/ydb-platform/ydb-java-dialects/pull/181/files#diff-d7f3eb1fd5295c12d964853abb1bfff142af528d4ac43039b105d55ac556b8d4R91).

🙈 `JdbcQueryLookupStrategy` is package-protected by design and to give us flexibility to change arguments without guaranteeing that it is a stable, public API.

> The idea is that the user must be able to dynamically provide the SQLType of the underlying parameter.

It sounds that this is the actual requirement. I wondered already whether we could come up with a mechanism to describe certain parameter aspects (e.g. in Cassandra we have [`@CassandraType`](https://github.com/spring-projects/spring-data-cassandra/blob/02987fe22ca3be46430e7e2e2614e4764b61f2b5/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/repository/isolated/RepositoryQueryMethodParameterTypesIntegrationTests.java#L191-L193) that can be used also on repository query method parameters). I could well imagine a proper `@SqlType("uint64")` approach and some Dialect-specific resolver that takes a type string and resolves it into `java.sql.SQLType`.

That would be IMO a much cleaner approach requiring less reliance on internals.

mipo256:
> Encapsulation and design. Not everything a constructor receives is intended to be exposed publicly. 

I'm not asking to make the accessors `public`, let's make them `protected`. 

> We hide certain aspects to avoid coupling (everything that is even remotely visible is going to be used. Everything being used and then changed generates tickets that ask for a reply).

I perfectly know that. The problem is that there is a balance. You can hide everything out, ad make the framework difficult to extend, and that would lead to fewer users. Or you can expose a lot, and then the framework is highly customizable and etc.  but changing any api results in a breaking change, and we go down that path. I understand it. My point is - there is a balance, and I honestly do not really think that making accessors protected is that much of burden in terms of backward compatibility. 

> It sounds that this is the actual requirement. I wondered already whether we could come up with a mechanism to describe certain parameter aspects (e.g. in Cassandra we have [@CassandraType](https://github.com/spring-projects/spring-data-cassandra/blob/02987fe22ca3be46430e7e2e2614e4764b61f2b5/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/repository/isolated/RepositoryQueryMethodParameterTypesIntegrationTests.java#L191-L193) that can be used also on repository query method parameters). I could well imagine a proper @SqlType("uint64") approach and some Dialect-specific resolver that takes a type string and resolves it into java.sql.SQLType.

The spring-data-cassandra [does exactly what we're doing, exactly that](https://github.com/spring-projects/spring-data-cassandra/blob/main/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/query/CassandraQueryMethod.java#L133). We can, of course, add a new annotation in the spring-data-jdbc module like `@JdbcType` and then consider it when resolving the `SQLType`.

mp911de:
Paging @schauder: What do you think about introducing a `@SqlType` parameter annotation in JDBC? Seems much cleaner than trying to extend from `JdbcQueryLookupStrategy`. In any case, those methods aren't required if you create subclasses and expose getters yourself so closing this PR without merge.

schauder:
An `@SqlType` sounds like a good thing to me.

mipo256:
Can I'll file an issue and the PR for this in Spring Data Relational? @schauder

schauder:
For the `@SqlType` stuff? Sure.

PR #3262, opened 2025-03-29
Fix broken docs caused by incorrect `feature-scroll` usage in Antora

dmitrysulman:
There are a couple of issues with the `feature-scroll` attribute in the Antora docs:

- The `endif::[]` for the _Offset-based Window<T>_ table row is placed incorrectly:
https://github.com/spring-projects/spring-data-commons/blob/f4e9581d23472fb40f45346ec1ed2e5cb0a878e0/src/main/antora/modules/ROOT/pages/repositories/query-methods-details.adoc?plain=1#L455-L464
This causes the table to break when `feature-scroll` [is not set](https://docs.spring.io/spring-data/commons/reference/repositories/query-methods-details.html#repositories.scrolling.guidance):
Screenshot 2025-03-29 at 22 02 53

- Defining `:feature-scroll: true` in [index.adoc](https://github.com/spring-projects/spring-data-commons/blob/f4e9581d23472fb40f45346ec1ed2e5cb0a878e0/src/main/antora/modules/ROOT/pages/index.adoc?plain=1#L6) has no effect because attributes set on one page are [not visible to other pages](https://docs.antora.org/antora/latest/playbook/asciidoc-attributes/#precedence-rules). Additionally, setting the attribute value to `true` is redundant — `ifdef` blocks are evaluated based on whether the attribute is defined, regardless of its value ([source](https://docs.asciidoctor.org/asciidoc/latest/directives/ifdef-ifndef/)).

**As a result, Scroll API-related content is always hidden in the documentation.**

This PR fixes the above by explicitly defining the `:feature-scroll:` attribute in the two pages where it's used:

- [query-methods-details.adoc](https://github.com/spring-projects/spring-data-commons/blob/main/src/main/antora/modules/ROOT/pages/repositories/query-methods-details.adoc)
- [core-concepts.adoc](https://github.com/spring-projects/spring-data-commons/blob/main/src/main/antora/modules/ROOT/pages/repositories/core-concepts.adoc)

### DIFF
--- a/src/main/antora/modules/ROOT/pages/index.adoc
+++ b/src/main/antora/modules/ROOT/pages/index.adoc
@@ -3,7 +3,6 @@
 Oliver Gierke; Thomas Darimont; Christoph Strobl; Mark Pollack; Thomas Risberg; Mark Paluch; Jay Bryant
 :revnumber: {version}
 :revdate: {localdate}
-:feature-scroll: true
 
 (C) 2008-2025 The original authors.
 
--- a/src/main/antora/modules/ROOT/pages/repositories/core-concepts.adoc
+++ b/src/main/antora/modules/ROOT/pages/repositories/core-concepts.adoc
@@ -1,5 +1,6 @@
 [[repositories.core-concepts]]
 = Core concepts
+:feature-scroll:
 
 The central interface in the Spring Data repository abstraction is `Repository`.
 It takes the domain class to manage as well as the identifier type of the domain class as type arguments.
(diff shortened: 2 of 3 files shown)


dmitrysulman:
Overriding this attribute for a specific `spring-data-` project can be done by [unsetting the attribute at the Antora playbook level](https://docs.antora.org/antora/latest/playbook/asciidoc-attributes/#hard-unset).

mp911de:
Thank you for your contribution. That's merged, polished, and backported now. In query-method-details, we rather want to keep opt-in as the majority of modules doesn't support scrolling (window) queries.

PR #3269, opened 2025-04-15
Tiny improvement of examples.

### LINKED ISSUE #3268: Documentation enhancement: Separate examples of methods with special parameters and methods without
[The documentation of Limiting Query Results](https://docs.spring.io/spring-data/jpa/reference/repositories/query-methods-details.html#repositories.limit-query-result) can be enhanced with the clarification and [comment from schauder](https://github.com/spring-projects/spring-data-jpa/issues/3831#issuecomment-2801840158).

Current situation:
The examples with parameters inside brackets such as `List<User> findByLastname(Limit limit);` and the examples without parameters inside brackets such as `User findFirstByOrderByLastnameAsc();` appear in the same box of examples.

Potential enhancement:
To be more obvious, it could be done like this:

I imagine a potential change could be: (There could be a better way, but this is what I can think of right now.)

Create two boxes of examples:
Box A:
Examples without parameters in brackets.

Box B:
Examples with parameters in brackets.

schauder:
Examples for limiting queries with and without a property argument are now better mixed.
This makes it more clear that limiting and filtering by property are orthogonal concerns.

Closes #3268

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-commons</artifactId>
-	<version>3.5.0-SNAPSHOT</version>
+	<version>3.5.0-tiny-sample-improvement-SNAPSHOT</version>
 
 	<name>Spring Data Core</name>
 	<description>Core Spring concepts underpinning every Spring Data module.</description>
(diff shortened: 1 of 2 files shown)


mp911de (approved):

schauder:
Merged and back ported.

PR #3272, opened 2025-04-18
Add `getRequiredName` and `hasName` API to `org.springframework.data.mapping.Parameter`.

### LINKED ISSUE #3088: Introduce `org.springframework.data.mapping.Parameter.getRequiredName()` and `Parameter.hasName()`
Provide a more convenient API to simplify the caller side especially for conditionals that want to determine whether a parameter name is present.

Typically, code checks for `null` and then throws an exception.

onobc:
Introduces a more convenient API to simplify the caller side especially for conditionals that want to determine whether a parameter name is present.

Closes #3088


mp911de:
That's merged, polished, and forward-ported now.

Issue #2937, opened 2023-09-19
ProxyingHandlerMethodArgumentResolver conflict with @AuthenticationPrincipal


spadou:
In a Spring application (with Spring Boot 3.1.3 currently), using data and security the `ProxyingHandlerMethodArgumentResolver` will conflict with the use of `@AuthenticationPrincipal` in a controller when a custom interface is used for the principal. The `ProxyingHandlerMethodArgumentResolver` will be registered before `AuthenticationPrincipalArgumentResolver` and handle the argument as a projection in the fallback preventing `@AuthenticationPrincipal` from working.

I see this was improved in #1237, but only the Spring packages are filtered out. In my case I use a custom interface for the principal to inject due to others requirements, so it is not ignored by the `ProxyingHandlerMethodArgumentResolver`. The list of ignored packages also doesn't seem configurable so it is not an option to handle the issue. I could also exclude the `SpringDataWebAutoConfiguration` but this exclude everything (like pageable) so it is not ideal.

There is already a `@ProjectedPayload` in `ProxyingHandlerMethodArgumentResolver` so wouldn't it make sense to remove the fallback as it could cause issues in multiple situations?

Here a small example that reproduce the issue:
```java
@SpringBootApplication
public class ArgumentResolverConflictApplication {

    public static void main(String[] args) {
        SpringApplication.run(ArgumentResolverConflictApplication.class, args);
    }

    @RestController
    public static class ArgumentResolverConflictRestController {

        @GetMapping("/principal")
        String principal(@AuthenticationPrincipal CustomUserDetails principal) {
            return principal.getUsername();
        }
    }

    @Bean
    public InMemoryUserDetailsManager inMemoryUserDetailsManager() {
        return new InMemoryUserDetailsManager(User.withUsername("user").password("{noop}password").build()) {
            @Override
            public UserDetails loadUserByUsername(String username) {
                var user = super.loadUserByUsername(username);
                return new CustomUser(user.getUsername(), user.getPassword(), user.isEnabled(), user.isAccountNonExpired(), user.isCredentialsNonExpired(),
                        user.isAccountNonLocked(), user.getAuthorities());
            }
        };
    }

    public interface CustomUserDetails extends UserDetails {
    }

    public static class CustomUser extends User implements CustomUserDetails {
        public CustomUser(String username, String password, boolean enabled, boolean accountNonExpired, boolean credentialsNonExpired, boolean accountNonLocked,
[...]

PR #3277, opened 2025-04-24
Avoid matching multipart parameters annotated with @ModelAttribute

### LINKED ISSUE #3258: Spring Data Commons 3.3.4 breaks `@ModelAttribute` handling
As explored in #33791 and visible with a slightly modified version of the reproducer https://github.com/EnyaEnya/multipart-null, adding `org.springframework.boot:spring-boot-starter-data-jpa` to the project breaks `@ModelAttribute` handling as of Spring Boot 3.3.4, and I have been able to reproduce only by adding `org.springframework.data:spring-data-commons:3.3.4` to the project. 

The same project with `org.springframework.data:spring-data-commons:3.3.3` or without Spring Data is fine.

onobc:
The ProxyHandlerMethodArgumentResolver now avoids matching multipart parameters annotated with @ModelAttribute. This allows multipart parameters to be handled by RequestParamMethodArgumentResolver which properly handles multipart arguments.

In 3.3.3 the `ProxyingHandlerMethodArgumentResolver` would report that it **does not** support the `@ModelAttribute MultipartFile file` parameter and then the next handler in the chain (`RequestParamMethodArgumentResolver`) was able to properly handle the multipart file arg.

In 3.3.4 ([commit](https://github.com/mipo256/spring-data-commons/commit/8ab55e4a195691202ec9fe8d510135443fed1353#diff-49193f625b12d1a6678bf4aa6b5a8eedd6edcbc3a208da6e8f4f8f8bfe302f09R32-R93)) the `ProxyingHandlerMethodArgumentResolver` reports that it **does** support the `@ModelAttribute MultipartFile file` parameter and then it fails to properly handle the multipart file arg.

Fixes #3258
Related tickets #2937

### DIFF
--- a/src/main/java/org/springframework/data/web/ProxyingHandlerMethodArgumentResolver.java
+++ b/src/main/java/org/springframework/data/web/ProxyingHandlerMethodArgumentResolver.java
@@ -36,11 +36,13 @@ import org.springframework.web.bind.support.WebDataBinderFactory;
 import org.springframework.web.context.request.NativeWebRequest;
 import org.springframework.web.method.annotation.ModelAttributeMethodProcessor;
 import org.springframework.web.method.support.HandlerMethodArgumentResolver;
+import org.springframework.web.multipart.support.MultipartResolutionDelegate;
 
 /**
  * {@link HandlerMethodArgumentResolver} to create Proxy instances for interface based controller method parameters.
  *
  * @author Oliver Gierke
+ * @author Chris Bono
  * @since 1.10
  */
 public class ProxyingHandlerMethodArgumentResolver extends ModelAttributeMethodProcessor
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/web/ProxyingHandlerMethodArgumentResolver.java
```diff
@@ -88,9 +90,9 @@ public boolean supportsParameter(MethodParameter parameter) {
 			return false;
 		}
 
-		// Annotated parameter
-		if (parameter.getParameterAnnotation(ProjectedPayload.class) != null
-				|| parameter.getParameterAnnotation(ModelAttribute.class) != null) {
+		// Annotated parameter (excluding multipart @ModelAttribute)
+		if (parameter.hasParameterAnnotation(ProjectedPayload.class) ||
```
onobc, line 94: It seems that the proxy handler does not properly handle multipart arguments, but the request param method handler does properly support multipart params. Is this the proper fix or should the proxy handler also support multipart args properly? cc: @odrotbohm @mp911de

mp911de: We have various aspects here. We should deprecate unannotated proxying as that has proven to cause quite some issues. We also need to have a plan how to achieve this and how our migration path would look like.

I'm thinking about:

* (backport) 3.4.x: Introduce a warning log if a parameter is not annotated with `@ProjectedPayload` that this style is deprecated and that we will drop support for projections if a parameter is not annotated with `@ProjectedPayload`
* 3.5: stop supporting parameters that are not annotated with `@ProjectedPayload` but keep the logging
* 4.0: remove logging

Introducing warn logging can cause quite some fallout, so we should be mindful about the conditions and maybe once we have seen a parameter, that we do not repeat logging (caching?) for that particular occurence.

onobc:
Oops, wrong rebase branch. Fixing now...

mp911de:

PR #3282, opened 2025-04-29
Refine Repository Composition retrieval during AOT.

christophstrobl:
Add module identifier and base repository implementation properties (seeks better alternative, maybe).
Fix fragment function previously overriding already set property due to name clash.
Extend tests for bean definition resolution and code block creation.


christophstrobl:
Merged to _4.0.x_.

PR #3285, opened 2025-05-05
Introduce `SearchResult` and `SearchResults`

mp911de:
We now provide value types representing a `SearchResult` similar to `GeoResult` and `GeoResults`.


christophstrobl:
merged to _4.0.x_

PR #3314, opened 2025-06-16
Polish README

ngocnhan-tran1996:
I've tried to build the documentation on my local machine, but it didn't work.

https://github.com/spring-projects/spring-data-commons?tab=readme-ov-file#building-reference-documentatio

It only worked when I added `antora:antora` at the end, and the generated documentation was available at `/target/site/index.html`.
I'm using Windows OS, so there might be differences compared to other OS.

Please check, and if it's correct, I will update it in this PR.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -73,7 +73,7 @@ https://github.com/spring-projects/spring-data-commons/issues[issue tracker] to 
 
 == Building from Source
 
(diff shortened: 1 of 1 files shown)


mp911de:
Thanks for your contribution. For the Antora build, we use the `-Pantora` profile. I've fixed the profile and the target path in a follow-up commit.

PR #3337, opened 2025-07-30
refactor: extract magic number 0L to constant for new entity check

currenjin:
### Motivation

Replaced the hardcoded value `0L` with a named constant to clarify its meaning as a marker for new (unsaved) entities.

Even though the value was only used in a single location, extracting it improves code readability and prevents potential misuse.

### Changes

- Introduced a private static final constant `NEW_ID = 0L`
- Replaced inline `0L` usage in new entity check

### Checklists


mp911de:
Thank you for your contribution. I'm not quite sure this is appropriate. We would have to add another constant for `null` if the identifier is an object type and not an primitive to reflect the entire story. Introducing constants isn't helpful as it doesn't help in explaining the concept. Also, we have other places (like `IsNewStrategy`) that follow a similar behavior.

Therefore, we do not want to merge this PR.

PR #3352, opened 2025-08-21
Example code in the scrolling document might not consume every element in the database.

r6hk:
The example code in the document: 

```java
Window<User> users = repository.findFirst10ByLastnameOrderByFirstname("Doe", ScrollPosition.offset());
do {

  for (User u : users) {
    // consume the user
  }

  // obtain the next Scroll
  users = repository.findFirst10ByLastnameOrderByFirstname("Doe", users.positionAt(users.size() - 1));
} while (!users.isEmpty() && users.hasNext());
```

The execution result is as follows (there are slight differences from the code in the documentation, but it illustrates the point):

screenshot-20250821-170027

It can be observed that the elements with indices 10 through 14 have not been consumed.

Modified code:

```java
ScrollPosition position = ScrollPosition.offset();
while (true) {
    users = repository.findFirst10ByLastnameOrderByFirstname("Doe", position);
    if (users.isEmpty()) break;
    for (User u : users) {
        // consume the user
    }
    if (!users.hasNext()) break;
    position = users.positionAt(users.size() - 1);
}
```

This code accounts for more cases, such as when the count of data is 0 or an exact multiple of 10.

### DIFF
--- a/src/main/antora/modules/ROOT/pages/repositories/scrolling.adoc
+++ b/src/main/antora/modules/ROOT/pages/repositories/scrolling.adoc
@@ -11,16 +11,16 @@ Similar to consuming a Java `Iterator<List<…>>` by obtaining the next batch of
 
 [source,java]
 ----
-Window<User> users = repository.findFirst10ByLastnameOrderByFirstname("Doe", ScrollPosition.offset());
-do {
-
+ScrollPosition position = ScrollPosition.offset();
+while (true) {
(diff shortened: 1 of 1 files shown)


mp911de:
Thanks for bringing this up. Actually, the condition is slightly off, it should be `||` instead of `&&`, so either the result has contents or the scrolling isn't finished yet. Do you want to include the change or should we just swap the operators?

PR #3363, opened 2025-09-16
Refine `TypeCollector` usage and predicates

mp911de:
Make predicates on `TypeCollector` usable, introduce strategy interface so that AOT participants can contribute their own predicates for inclusion/exclusion of types that should be considered reachable.

### DIFF
--- a/src/main/java/org/springframework/data/util/TypeCollector.java
+++ b/src/main/java/org/springframework/data/util/TypeCollector.java
@@ -109,7 +109,7 @@ public class TypeCollector {
 	 */
 	@Contract("_ -> this")
 	public TypeCollector filterMethods(Predicate<Method> filter) {
-		this.methodFilter = filter.and(filter);
+		this.methodFilter = methodFilter.and(filter);
 		return this;
 	}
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/data/util/TypeCollector.java
```diff
@@ -275,5 +334,111 @@ public boolean isEmpty() {
...
+	 * {@value AotServices#FACTORIES_RESOURCE_LOCATION} resource. This interface serves as SPI and can be provided through
+	 * {@link org.springframework.beans.factory.aot.AotServices}.
+	 * <p>
+	 * {@link TypeCollector} discovers all implementations and applies the combined predicates returned by this interface
+	 * to filter unwanted reachable types from AOT contribution.
+	 *
+	 * @author Mark Paluch
+	 * @since 4.0
+	 */
+	public interface TypeCollectorPredicateProvider {
```
mp911de, line 354: What if we renamed this to `TypeCollectorPredicates`?

christophstrobl: a bit confusing, naming it predicates, `TypeCollectorFilterProvider` maybe?

PR #3371, opened 2025-09-25
Use `StableValue` behind `Lazy`

mp911de:
**Update:** JEP 502's successor is a draft for [`LazyConstants`](https://cr.openjdk.org/~pminborg/lazy-constants-second-preview/api/java.base/java/lang/LazyConstant.html) https://openjdk.org/jeps/8359894 with a much more aligned API.

We now use Java's `StableValue` to back our `Lazy` implementation. `StableValue` suggests an Optional orElse…(…) usage pattern and it is not primarily intended for lazifying value resolution, therefore we have to mimic `StableValue.supplier(…)`.

An immediate benefit is that Java 25++ users will get concurrency guarding while our `Lazy` implementation uses a best effort non-locking approach for value resolution that may observe race conditions as we don't guarantee single `Supplier` access for the backing supplier object.

We're using Java's preview features through a multi-release jar keeping our Java 17 baseline but enabling [Java 25 `StableValue`](https://openjdk.org/jeps/502) feature consumption. Using the `4.0.0-STABLE-VALUE-SNAPSHOT` version requires enabling Java preview features during compilation and runtime.

```xml
<dependencies>
    <dependency>
        <groupId>org.springframework.data</groupId>
        <artifactId>spring-data-commons</artifactId>
        <version>4.0.0-STABLE-VALUE-SNAPSHOT</version>
    </dependency>
</dependencies>

<repositories>
    <repository>
        <id>spring-snapshot</id>
        <url>https://repo.spring.io/snapshot</url>
        <snapshots>
            <enabled>true</enabled>
        </snapshots>
    </repository>
</repositories>
```

### DIFF
--- a/ci/pipeline.properties
+++ b/ci/pipeline.properties
@@ -1,10 +1,10 @@
 # Java versions
-java.main.tag=24.0.1_9-jdk-noble
-java.next.tag=24.0.1_9-jdk-noble
+java.main.tag=25
+java.next.tag=25-cds
 
 # Docker container images - standard
-docker.java.main.image=library/eclipse-temurin:${java.main.tag}
-docker.java.next.image=library/eclipse-temurin:${java.next.tag}
+docker.java.main.image=bellsoft/liberica-openjdk-alpine:${java.main.tag}
+docker.java.next.image=bellsoft/liberica-openjdk-alpine:${java.next.tag}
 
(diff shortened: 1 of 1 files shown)


mp911de:
Will be superseded by `LazyConstant`.

Issue #3375, opened 2025-10-06
Improve `PropertyMatch` performance using precompiled regex patterns


kamilkrzywanski:
**Current behavior**

In `org.springframework.data.mapping.PropertyMatch`, the `matches(String name, Class<?> type)` method currently uses `String.matches(...)`:

```
if ((namePattern != null) && !name.matches(namePattern)) {
    return false;
}

```
I noticed this while profiling application startup — PropertyMatch.matches(...) is invoked repeatedly during Spring Data repository and mapping initialization, which makes the repeated regex compilation visible in startup.

Proposed improvement
Pre-compile the namePattern in the constructor of PropertyMatch as a java.util.regex.Pattern.
Change matches(...) to use Pattern.matcher(...).matches().

PR #3376, opened 2025-10-06
Use precompiled regex for `PropertyMatch.namePattern`

kamilkrzywanski:
Replaced String.matches(..) with a precompiled Pattern to avoid repeated regex compilation and improve startup performance. Issue: https://github.com/spring-projects/spring-data-commons/issues/3375


mp911de:

PR #3384, opened 2025-10-21
Update documentation

christophstrobl:
Do not merge right away - rather generic to capture various documentation changes.

### DIFF
--- a/src/main/antora/modules/ROOT/pages/auditing.adoc
+++ b/src/main/antora/modules/ROOT/pages/auditing.adoc
@@ -34,6 +34,8 @@ class Customer {
 As you can see, the annotations can be applied selectively, depending on which information you want to capture.
 The annotations, indicating to capture when changes are made, can be used on properties of type JDK8 date and time types, `long`, `Long`, and legacy Java `Date` and `Calendar`.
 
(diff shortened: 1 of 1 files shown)


src/main/antora/modules/ROOT/pages/aot.adoc
```diff
@@ -32,30 +37,110 @@ However, there users may fine tune the configuration with following options.
 
 |`spring.aot.data.accessors.include`
 |Comma separated list of FQCN for which to contribute Bytecode for generated Type/Property Accessors.
-Ant-style include patterns matching package names (e.g. `com.acme.**`) or type names inclusion.
+Ant-style include patterns matching package names (e.g. `example.springdata.**`) or type names inclusion.
 Inclusion pattern matches are evaluated before exclusions for broad exclusion and selective inclusion.
```
schauder: I find this order of evaluation rather confusing, because it is really about precedence. 

Something like this seems easier to understand.

If a type is matched by a inclusion and an exclusion, the exclusion _wins_ and the type is considered excluded.

src/main/antora/modules/ROOT/pages/aot.adoc
```diff
@@ -32,30 +37,110 @@ However, there users may fine tune the configuration with following options.
...
 
 |`spring.aot.data.accessors.exclude`
 |Comma separated list of FQCN for which to skip contribution of Bytecode for generated Type/Property Accessors.
-Ant-style exclude patterns matching package names (e.g. `com.acme.**`) or type names exclusion.
+Ant-style exclude patterns matching package names (e.g. `example.springdata.**`) or type names exclusion.
 Exclusion pattern matches are evaluated after inclusions for broad exclusion and selective inclusion.
```
schauder: See above.

Issue #3177, opened 2024-10-15
Allow overriding DateTimeProvider via Bean Definition


ZIRAKrezovic:
The default DateTimeProvider is set to CurrentDateTimeProvider.INSTANCE which returns LocalDateTime.now().

We have to override it when using OffsetDateTime as JPA will throw an error.

According to https://docs.spring.io/spring-data/jpa/reference/auditing.html#jpa.auditing.configuration it is already possible to override AuditorAware by simply defining a bean.

However, the same is not true for DateTimeProvider. I have tried defining a bean of type DateTimeProvider, but it was not picked up. I was only able to specify it via `@EnableJpaAuditing(dateTimeProviderRef = "...")`

(note, I used EnableJpaAuditing, but AuditingHandlerSupport is part of spring-data-commons)

mp911de:
`AuditingBeanDefinitionRegistrarSupport` enables autowire by type if `auditorAwareRef` isn't set. I think we should be able to generally set autowrite mode to `AUTOWIRE_BY_TYPE` and back off from setting `dateTimeProvider` to `CurrentDateTimeProvider.INSTANCE` if no reference is configured as `AuditingHandlerSupport` initializes itself with `CurrentDateTimeProvider.INSTANCE`.

PR #3385, opened 2025-10-21
Allow overriding DateTimeProvider via Bean Definition

ghgh415263:
use autowired properties for AuditorAware and DateTimeProvider when explicit references are not provided

See: #3177


ghgh415263:
I added unit tests.
I would appreciate any feedback if you notice any issues.

PR #3409, opened 2025-11-17
Introduce first-class type-safe property path references at the core level

mp911de:
We now support type-safe property paths and property references:

Java variants:

```java
PropertyPath.from("name", Person.class) // existing String-based API
PropertyPath.of(Person::getName) // type-safe property reference expression

PropertyPath.from("address.country", Person.class) // existing nested path API
PropertyPath.of(Person::getAddress).then(Address::getCountry) // type-safe composed path expression


PropertyReference.of(Secret::getSecret)
```

Kotlin variants:

```kotlin
PropertyReference.of(Secret::secret)

PropertyPath.of(Person::address / Address::city)
```

allowing type-safe usage through e.g.:

```java
Sort.by(Person::getFirstName, Person::getLastName)
```

Module adoption:

- [x] https://github.com/spring-projects/spring-data-jpa/pull/4086
- [x] https://github.com/spring-projects/spring-data-mongodb/issues/5135
- [x] https://github.com/spring-projects/spring-data-relational/issues/2226
- [x] https://github.com/spring-projects/spring-data-cassandra/pull/1626

PR #3446, opened 2026-01-21
Add kotlin extensions for `Sort`.

### LINKED ISSUE #3445: Add Kotlin extension for `Sort`
I will create PR for this issue.

current:
```java
Query()
    .with(Sort.by(Sort.Direction.DESC, WebhookDelivery::createdAt.name))
```

with extension:
```kt
Query()
    .with(WebhookDelivery::createdAt.desc())
```

nillpoe:

Closes #3445


nillpoe:
Hi, thanks for the reply. I missed that one because the search for "sort" didn't yield it.
Yes, I think that is another form of Kotlin typed-property support.

But this change will allow you to choose between ascending and descending more easily in most use case.
So I don't really think #3409 closes this PR.

PR #3453, opened 2026-02-02
Stop supporting params not annotated with `@ProjectedPayload`

### LINKED ISSUE #3301: Stop supporting params that are not annotated with `@ProjectedPayload` (4.1.x)
This stops supporting parameters that are not annotated with `@ProjectedPayload` but keep the logging (introduced in #3300) , adjusting the message accordingly. 

Also, add this detail to "Removals/Breaking Changes" section of docs/release notes.

> [!IMPORTANT]
> This is intended to be done in the 4.1.x line

> [!NOTE]
> **Implementation notes:**
> - Avoid log spamming: only log the warning once per param (or maybe drop down from warn to debug)
> - Allow users to disable the logging via the typically logging config

onobc:
This commit drops support for projections if a parameter (or the parameter type) is not explicitly annotated with `@ ProjectedPayload`. A warning is still logged in these cases to help user understand why the parameter is not being projected.

Closes #3301

### DIFF
--- a/src/main/java/org/springframework/data/web/ProxyingHandlerMethodArgumentResolver.java
+++ b/src/main/java/org/springframework/data/web/ProxyingHandlerMethodArgumentResolver.java
@@ -115,7 +115,7 @@ public class ProxyingHandlerMethodArgumentResolver extends ModelAttributeMethodP
 		// Parameter annotated with @ModelAttribute
 		if (parameter.hasParameterAnnotation(ModelAttribute.class)) {
 			this.deprecationLogger.logDeprecationForParameter(parameter);
-			return true;
+			return false;
 		}
 
 		// Exclude any other parameters annotated with Spring annotation
@@ -131,7 +131,7 @@ public class ProxyingHandlerMethodArgumentResolver extends ModelAttributeMethodP
 		String packageName = ClassUtils.getPackageName(type);
 		if (IGNORED_PACKAGES.stream().noneMatch(packageName::startsWith)) {
 			this.deprecationLogger.logDeprecationForParameter(parameter);
-			return true;
+			return false;
 		}
 
 		return false;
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/web/ProxyingHandlerMethodArgumentResolver.java
```diff
@@ -115,7 +115,7 @@ public boolean supportsParameter(MethodParameter parameter) {
 		// Parameter annotated with @ModelAttribute
 		if (parameter.hasParameterAnnotation(ModelAttribute.class)) {
 			this.deprecationLogger.logDeprecationForParameter(parameter);
```
onobc: In 4.2 when we rip out the logging we will simplify this logic greatly. Leaving as-is for now so that logging remains in the proper places.

src/main/java/org/springframework/data/web/ProxyingHandlerMethodArgumentResolver.java
```diff
@@ -157,7 +157,7 @@ protected void bindRequestParameters(WebDataBinder binder, NativeWebRequest requ
 	 */
 	static class ProjectedPayloadDeprecationLogger {
 
-		private static final String MESSAGE = "Parameter %sat index %s in [%s] is not annotated with @ProjectedPayload. Support for parameters not annotated with @ProjectedPayload (at the parameter or type level) will be dropped in a future version.";
+		private static final String MESSAGE = "Parameter %sat index %s in [%s] is not annotated with @ProjectedPayload. Parameters not annotated with @ProjectedPayload (at the parameter or type level) is not supported.";
```
mp911de: How about turning the sentence into a simpler form by removing the two negations? Like `… Make sure to annotate … with @ProjectedPayload to use it for projections` (something along those lines).

mp911de:

PR #3488, opened 2026-06-01
fix 'exits' to 'exists' in default methods section

barneyb:
In the default methods section, fix a reference to the `exits` [sic] default methods, to say `exists`.


christophstrobl:
Thank you @barneyb for spotting this - yet it has been discovered before. Closing as duplicate of #3482

PR #3506, opened 2026-07-05
Convert elements of collection- and array-valued query method parameters

### LINKED ISSUE #3502: Query exposed through Spring-Data-Rest will throw an exception if configured a filter on a List of enums
The following line of class `ReflectionRepositoryInvoker`
`result[i] = targetType.isInstance(value) ? value : convert(value, param);`
In case of targetType `List` doesn't verify if the generic element required is an enum or a String, skipping the conversion into List of enum.
Hibernate will throw an exception if we provide at least 2 elements.
The convert method, skipped due to the true condition, would have handled the conversion avoiding the error

https://github.com/spring-projects/spring-data-commons/blob/0d3cf4ab81a071551631461a7933bd501f14d23a/src/main/java/org/springframework/data/repository/support/ReflectionRepositoryInvoker.java#L191

I reproduced the error on the following repository:
https://github.com/antonioT90/spring-data-rest-enums-errors-sample

seonwooj0810:
Fixes #3502

**Root cause**
`ReflectionRepositoryInvoker#prepareParameters` skipped conversion whenever the resolved value already was an instance of the declared parameter type. For a collection- or array-valued parameter this is wrong: a repeated query parameter (e.g. `?colors=RED&colors=GREEN`) resolves to a `List<String>`, which *is* an instance of the target `List`/`Collection`, so `convert(...)` was never called and the raw `String` elements were never converted to the declared element type (e.g. an enum). The unconverted `List<String>` then reached the `in`-style query and the persistence provider failed. (As diagnosed by @antonioT90 in the issue, with a reproducer.)

**Change**
For array- and `Iterable`-typed parameters, conversion is now performed even when the value already matches the target type, so the elements are converted to the declared element type. Scalar parameters keep the previous fast-path (no redundant conversion).

**Tests**
Added `ReflectionRepositoryInvokerUnitTests#convertsElementsOfCollectionValuedQueryParameter`, which binds a `List<String>` to a `Collection<Color>` query parameter and asserts the repository receives converted enum values. It fails before the change (repository receives the raw `String`s) and passes after.

Verification done: (1) no in-flight PR — searched open PRs for `3502` and `ReflectionRepositoryInvoker` (only unrelated #3439); (2) no self-claim comments; assignee `odrotbohm` is routine triage tracking (89 open issues assigned); (3) code-focused, touches only `.java`; (4) confirmed the pattern still present on current `main`; (6) issue is past triage (`type: bug`), so no triage gate. Ran `./mvnw -o test -Dtest=ReflectionRepositoryInvokerUnitTests` — 20/20 pass with the fix; the new test fails without it.


seonwooj0810:
Superseded by #3514 — recreated from the same branch after this PR was inadvertently closed by a force-push on my fork. Same change, with the commit now signed off (DCO).

PR #3510, opened 2026-07-10
Support legacy Kotlin property path rendering.

### LINKED ISSUE #3503: Kotlin `org.springframework.data.core.div` extension drops parent and intemediate path segments
Hi Spring Data team !

Following my analysis in spring-projects/spring-data-mongodb#5216 on Kotlin DSL behaviors. 

Currently, when using the Kotlin DSL for mongodb, we can use the `div` operator to query nested properties:

```kotlin
import org.springframework.data.mapping.div

mongoOperations.find<Book>(
  Query(Book::author / Author::name regex "^H")          
)
// { "author.name": /^H/ }
```

However, the `div` operator is now deprecated, with a recommendation to use the one from `org.springframework.data.core.div`. At the moment, if we replace the `div` operator with the new one, it generates the following query: `{ "name": /^H/ }`

```kotlin
import org.springframework.data.core.div

mongoOperations.find<Book>(
  Query(Book::author / Author::name regex "^H")          
)
// { "name": /^H/ }
```

This behavior is incorrect, as it drops the parent path and may lead to invalid queries.

IMHO, it would be a good idea to extend the `org.springframework.data.core.asString(property: KProperty<*>)` function to support the previous `org.springframework.data.mapping.KPropertyPath` and `org.springframework.data.mapping.KIterablePropertyPath` types within the when statement. (see [asString](https://github.com/spring-projects/spring-data-commons/blob/3c60951c92900cd538ee41d6196f63a6ddd5d308/src/main/kotlin/org/springframework/data/core/KPropertyPath.kt#L106) )

This would ensure that calls made through `org.springframework.data.core.toDotPath()` correctly handle both versions of the `div` operator.

Updating spring-data-mongodb to rely on `org.springframework.data.core.toDotPath()` would allow users to use either `org.springframework.data.core.div` or `org.springframework.data.mapping.div` while producing identical queries.

I think this would provide a smoother migration path for users.

Any thoughts ?

Regards,
Patouche

hutiefang76:
The Kotlin property-path DSL moved from `org.springframework.data.mapping` to `org.springframework.data.core`, but a path built with the deprecated mapping `div` extensions was not recognized by the new core `toDotPath`. As a result, paths such as `author.name` and `books.author.name` were reduced to their final segment.

This change lets the core renderer recognize both legacy mapping path wrappers. It only expands the legacy non-iterable wrapper from `private` to `internal`; public extension signatures remain unchanged.

Tests:
- `./mvnw -Dtest=KPropertyPathTests test` (13 tests)
- Core and mapping property-path suites together (25 tests)
- `./mvnw -DskipTests package`
- `git diff --check`

- [x] I have read the Spring Data contribution guidelines.
- [x] I followed the existing Kotlin import and formatting style.
- [x] I added regression coverage for normal, iterable, and recursively nested legacy paths.
- [x] I added myself as an author to the Kotlin types and test class I touched.

Closes #3503

### DIFF
--- a/src/main/kotlin/org/springframework/data/core/KPropertyPath.kt
+++ b/src/main/kotlin/org/springframework/data/core/KPropertyPath.kt
@@ -17,6 +17,8 @@
 
 package org.springframework.data.core
 
+import org.springframework.data.mapping.KIterablePropertyPath as MappingKIterablePropertyPath
+import org.springframework.data.mapping.KPropertyPath as MappingKPropertyPath
 import kotlin.reflect.KProperty
 import kotlin.reflect.KProperty1
 
@@ -101,6 +103,7 @@ internal class KIterablePropertyReference<T, M, out P>(
  *
  * @author Tjeu Kayim
  * @author Mikhail Polivakha
+ * @author hutiefang
  * @since 4.1
  */
 internal fun asString(property: KProperty<*>): String {
@@ -109,8 +112,13 @@ internal fun asString(property: KProperty<*>): String {
 		is KPropertyPath<*, *> ->
 			"${asString(property.property)}.${property.leaf.name}"
 
+		is MappingKPropertyPath<*, *> ->
+			"${asString(property.parent)}.${property.child.name}"
+
+		is MappingKIterablePropertyPath<*, *> ->
+			"${asString(property.parent)}.${property.child.name}"
+
 		else -> property.name
 	}
 
 }
-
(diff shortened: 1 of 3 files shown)


mp911de:
We're not going to merge this pull request as it is not correctly addressing the entire issue. Depending on the property path renderer and used API, core -> mapping and mapping -> core div extensions can drop between the first and the intermediate segments silently.

PR #3511, opened 2026-07-11
Do not render meta-annotations for AOT repository method parameters

### LINKED ISSUE #3499: AOT repository generation emits invalid @Constraint annotation alongside jakarta.validation.constraints.* parameters
In our Spring Boot project, we do have a Spring AOT configuration that used to work with 4.0.3, but started failing after 4.0.6.

Example of repository: 
```java
import jakarta.validation.constraints.NotNull;
import java.util.List;
import java.util.UUID;
import org.springframework.data.jpa.repository.JpaRepository;
public interface BananaRepository extends JpaRepository<Banana, UUID> {
    List<Banana> findAllByStoreUuid(
        @NotNull UUID storeUuid
    );
}
```

Error during maven `process-aot`: 
`[ERROR] annotation @jakarta.validation.Constraint is missing a default value for the element 'validatedBy' `

To reproduce, please run: `mvn -B -q package -DskipTests` with [spring-repository-aot-issue.zip](https://github.com/user-attachments/files/28776046/spring-repository-aot-issue.zip)

We can argue about the usage of the validation annotation, but still, I believe this should be supported.
For now, we just removed them to fix the problem. 

It might relates to https://github.com/spring-projects/spring-data-commons/commit/852d789de7

seonwooj0810:
Fixes #3499

### Root cause
`MethodMetadata.buildParameter(…)` (added in 852d789 to render query-method parameter annotations for AOT repositories) iterates `MergedAnnotations.from(methodParameter.getParameterAnnotations())`. `MergedAnnotations` also exposes *meta*-annotations, so for a parameter annotated with e.g. `jakarta.validation.constraints.@NotNull`, its meta-annotation `@jakarta.validation.Constraint` was rendered standalone into the generated repository source. Because `@Constraint` has a mandatory `validatedBy` element, `process-aot` / native builds fail to compile the generated code:

```
annotation @jakarta.validation.Constraint is missing a default value for the element 'validatedBy'
```

This is a regression: repositories with `jakarta.validation.constraints.*` parameters that built with 4.0.3 broke on 4.0.6.

### Change
Only render **directly-present** annotations (`MergedAnnotation#isDirectlyPresent()`). Meta-annotations remain discoverable through the concrete annotation itself, so the introspection intent of #3458 is preserved while the invalid standalone meta-annotation is no longer emitted.

### Tests
Added `MethodMetadataUnitTests#doesNotRenderMetaAnnotations`, which declares a parameter annotated with a meta-annotated annotation (mirroring how `@NotNull` is meta-annotated with `@Constraint`) and asserts only the directly-present annotation is rendered. The test fails before the change and passes after.

### Verification done
- Reproduced the meta-annotation expansion in `buildParameter` on current `main`.
- New regression test fails on unpatched `main`, passes with the fix.
- Re-ran `MethodMetadataUnitTests`, `AotRepositoryMethodBuilderUnitTests`, `MethodContributorUnitTests`, `RepositoryContributorUnitTests` — all green.

### DIFF
--- a/src/main/java/org/springframework/data/repository/aot/generate/MethodMetadata.java
+++ b/src/main/java/org/springframework/data/repository/aot/generate/MethodMetadata.java
@@ -108,7 +108,9 @@ class MethodMetadata {
 		MergedAnnotations annotations = MergedAnnotations.from(methodParameter.getParameterAnnotations());
 
 		for (MergedAnnotation<Annotation> annotation : annotations) {
-			builder.addAnnotation(AnnotationSpec.get(annotation.synthesize()));
+			if (annotation.isDirectlyPresent()) {
+				builder.addAnnotation(AnnotationSpec.get(annotation.synthesize()));
+			}
 		}
 
 		return builder.build();
--- a/src/test/java/org/springframework/data/repository/aot/generate/MethodMetadataUnitTests.java
+++ b/src/test/java/org/springframework/data/repository/aot/generate/MethodMetadataUnitTests.java
@@ -18,6 +18,10 @@ package org.springframework.data.repository.aot.generate;
 import static org.assertj.core.api.Assertions.*;
 import static org.mockito.ArgumentMatchers.*;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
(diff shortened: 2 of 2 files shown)


seonwooj0810:
Checking in on this whenever you have a moment — happy to add more coverage if useful.

PR #3526, opened 2026-08-13
Document Vavr collection types as query method parameters

### LINKED ISSUE #3525: Support for Vavr Collections doesn't mention usage as repository query method parameters
## TL;DR
The **“Support for Vavr Collections”** docs only mention Vavr types as **query method return types**, which can misleadingly imply they are not supported as parameters.
However, Spring Data Commons automatically unwraps Vavr collection parameters to Java collections before Hibernate sees them. Methods such as `findByStatusIn(io.vavr.collection.List<Status>)` therefore work correctly with `IN` queries.

This was verified end-to-end with Spring Data JPA 3.5.13, Hibernate 6.6.53.Final, and Vavr 1.0.1.

**Request:** Please update the documentation to explicitly mention that Vavr collection types are also supported as **query method parameters**, not only return types.

## Full Context

The reference docs for [Support for Vavr Collections](https://docs.spring.io/spring-data/commons/reference/repositories/query-methods-details.html#repositories.collections-and-iterables.vavr) only describe Vavr (`io.vavr`) collection/`Option` types as query method **return** types, and Hibernate's own parameter-binding code only checks `value instanceof Collection` ([`QueryParameterBindingImpl#handleAsMultiValue`](https://github.com/hibernate/hibernate-orm/blob/6.6.53/hibernate-core/src/main/java/org/hibernate/query/internal/QueryParameterBindingImpl.java#L172-L185)), with no Vavr-awareness — combined, it's reasonable to conclude Vavr types aren't supported as method **parameters** at all.
This isn't just a theoretical risk: reading those two facts together led to exactly this conclusion during an actual internal code review (unrelated to Spring Data), flagged as a **correctness/runtime-crash** issue — the claim was that a derived query method taking a Vavr collection as an `IN`-clause argument would throw `IllegalArgumentException` at runtime, silently breaking a bulk `UPDATE` in production and leaving stale rows behind.

That conclusion doesn't hold: Spring Data Commons unwraps Vavr collection/`Option` arguments into their Java-native equivalent before the value ever reaches Hibernate, on every method argument, not just return values:

- [`ParametersParameterAccessor`](https://github.com/spring-projects/spring-data-commons/blob/3.5.13/src/main/java/org/springframework/data/repository/query/ParametersParameterAccessor.java#L47-L77) — constructor runs `QueryExecutionConverters.unwrap(...)` on every method argument whose type is a registered wrapper type.
- [`QueryExecutionConverters`](https://github.com/spring-projects/spring-data-commons/blob/3.5.13/src/main/java/org/springframework/data/repository/util/QueryExecutionConverters.java#L107-L143) — registers `io.vavr.collection.Seq`/`Map`/`Set` as wrapper types "as of 2.0" via `CustomCollections`.
- [`CustomCollections`](https://github.com/spring-projects/spring-data-commons/blob/3.5.13/src/main/java/org/springframework/data/util/CustomCollections.java#L358-L378) — `VavrToJavaCollectionConverter#convert` calls `((Seq<?>) source).asJava()` to hand back a real `java.util.List`/`Map`/`Set`.
- [`JpaParametersParameterAccessor`](https://github.com/spring-projects/spring-data-jpa/blob/3.5.13/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/JpaParametersParameterAccessor.java#L29) — extends `ParametersParameterAccessor` directly, so JPA inherits this unwrapping for free.
- [`ParameterMetadataProvider`](https://github.com/spring-projects/spring-data-jpa/blob/3.5.13/spring-data-jpa/src/main/java/org/springframework/data/jpa/repository/query/ParameterMetadataProvider.java#L83) — pulls bind values via `accessor.iterator()`, i.e., the already-unwrapped values, before Hibernate ever sees them.

This chain is unchanged across `2.7.0`, `3.3.0`, `3.5.13`, and current `main`, and predates Vavr support itself — the constructor-level unwrap was added in [commit `1185e71`](https://github.com/spring-projects/spring-data-commons/commit/1185e71214b774eca437bb0bfbf07fade9aee903) (2015, DATACMNS-768) for `java.util.Optional`/Guava `Optional`; Vavr collections inherited it for free once registered in the same `QueryExecutionConverters` registry over a year later.

Minimal reproduction: on `spring-data-jpa:3.5.13` / `hibernate-core:6.6.53.Final` / `vavr:1.0.1` *(or just **Spring Boot 3.5.16** with **Spring Data JPA** and **Vavr 1.0.1**)*, a derived query method declared as `findByStatusIn(io.vavr.collection.List<Status> activeStatuses)` feeding an `IN`-clause bulk `UPDATE` was exercised end-to-end against a live database (verified twice, with fresh row-level before/after evidence each time) and filtered/updated rows correctly, with no exception.

The story below walks through how that played out end-to-end, and the lesson it left behind.

<details>
<summary><h2>Story from that actual code review (click to expand)<h2></summary>

During review of a derived query method along the lines of `findByStatusIn(io.vavr.collection.List<Status> activeStatuses)` feeding an `IN`-clause bulk `UPDATE`, a reviewer flagged it as a **correctness/runtime-crash** issue: `io.vavr.collection.List` isn't a `java.util.Collection`, Hibernate's JPQL `IN`-clause binder requires one, and — per this very doc page — Vavr's Spring Data integration "covers return-type wrapping only."
Conclusion: every call would throw `IllegalArgumentException`, the bulk update would silently never run, and stale rows would pile up in production.

That objection turned out to be wrong, but only after real work to prove it: verifying against the exact dependency versions in use (`spring-data-jpa`, `hibernate-core`, `vavr`) via live testing — hitting the running application twice and inspecting the database directly before and after each call, with fresh row-level evidence each time — plus re-tracing the exact source chain described above (`ParametersParameterAccessor` → `QueryExecutionConverters` → the Vavr-to-Java collection converter) to show *why* the observed behavior was correct, not a fluke.

**The moral:** none of that legwork should have been necessary.
The two facts driving the objection — this doc page, and Hibernate's `instanceof Collection` check — are both public and both correct in isolation; the doc gap is the only reason they add up to a wrong conclusion.
Documenting the parameter-side behavior turns a multi-hour/day, evidence-gathering review debate into a one-line link.

</details>

---

Could the "Support for Vavr Collections" section (source: [`query-methods-details.adoc`](https://github.com/spring-projects/spring-data-commons/blob/3.4.1/src/main/antora/modules/ROOT/pages/repositories/query-methods-details.adoc#support-for-vavr-collections)) mention that these wrapper types are also accepted as query method **parameters**, not just return types?
That would save the next reader from having to trace through `QueryExecutionConverters`/`ParametersParameterAccessor` to confirm it, and prevent teams from concluding Vavr support is incomplete when it's actually just undocumented.

The same gap is duplicated in the [Supported Query Return Types](https://docs.spring.io/spring-data/jpa/reference/repositories/query-return-types-reference.html#appendix.query.return.types) appendix (source: [`query-return-types-reference.adoc`](https://github.com/spring-projects/spring-data-commons/blob/main/src/main/antora/modules/ROOT/pages/repositories/query-return-types-reference.adoc), same repo - verified byte-identical to the copy rendered under `spring-data-jpa`'s reference docs).
Its row for Vavr collection types reads "Vavr collection types. See [Support for Vavr Collections] for details.", framing them as a return type only, same as the section above.
Could that row also note that these types are accepted as method parameters?

(Searched existing issues for "vavr" + "parameter"/"argument" — didn't find a prior report of this specific documentation gap.)

AzazelSensei:
The Vavr section only talked about return types. Those wrappers are also unwrapped on the way in (ParametersParameterAccessor / QueryExecutionConverters), so findByStatusIn(io.vavr.collection.List) is a regular IN clause.

Fixes #3525

### DIFF
--- a/src/main/antora/modules/ROOT/pages/repositories/query-methods-details.adoc
+++ b/src/main/antora/modules/ROOT/pages/repositories/query-methods-details.adoc
@@ -280,9 +280,9 @@ You can use the types in the first column (or subtypes thereof) as query method 
 Alternatively, you can declare `Traversable` (the Vavr `Iterable` equivalent), and we then derive the implementation class from the actual return value.
 That is, a `java.util.List` is turned into a Vavr `List` or `Seq`, a `java.util.Set` becomes a Vavr `LinkedHashSet` `Set`, and so on.
 
-The same types are also accepted as query method *parameters*.
-Spring Data unwraps a Vavr `Seq`, `Set`, or `Map` argument to a Java collection before the store sees it, so a derived query such as `findByStatusIn(io.vavr.collection.List<Status> statuses)` binds as a regular `IN` clause.
-A Vavr `Option` argument is unwrapped to its value (or `null`), not to `java.util.Optional`.
+These Vavr types are also accepted as query method *parameters* (as of 2.0). If a repository method parameter uses one of the types above (or `Traversable`), Spring Data automatically unwraps it into its Java-native equivalent before executing the query.
+
+For example, a derived query method declared as `findAllByIdIn(io.vavr.collection.List<Long> ids)` accepts a Vavr `List` and unwraps it to a `java.util.List` so it binds correctly as an `IN`-clause. A Vavr `Option` parameter is similarly unwrapped to its inner value (or `null`).
 
 
 [[repositories.query-streaming]]
--- a/src/main/antora/modules/ROOT/pages/repositories/query-methods-details.adoc
+++ b/src/main/antora/modules/ROOT/pages/repositories/query-methods-details.adoc
@@ -280,6 +280,10 @@ You can use the types in the first column (or subtypes thereof) as query method 
 Alternatively, you can declare `Traversable` (the Vavr `Iterable` equivalent), and we then derive the implementation class from the actual return value.
 That is, a `java.util.List` is turned into a Vavr `List` or `Seq`, a `java.util.Set` becomes a Vavr `LinkedHashSet` `Set`, and so on.
 
+The same types are also accepted as query method *parameters*.
+Spring Data unwraps a Vavr `Seq`, `Set`, or `Map` argument to a Java collection before the store sees it, so a derived query such as `findByStatusIn(io.vavr.collection.List<Status> statuses)` binds as a regular `IN` clause.
+A Vavr `Option` argument is unwrapped to its value (or `null`), not to `java.util.Optional`.
+
 
 [[repositories.query-streaming]]
 == Streaming Query Results
--- a/src/main/antora/modules/ROOT/pages/repositories/query-return-types-reference.adoc
+++ b/src/main/antora/modules/ROOT/pages/repositories/query-return-types-reference.adoc
@@ -27,7 +27,7 @@ Some store modules may define their own result wrapper types.
 |`Stream<T>`|A Java `Stream`.
 |`Streamable<T>`|A convenience extension of `Iterable` that directly exposes methods to stream, map and filter results, concatenate them etc.
 |Types that implement `Streamable` and take a `Streamable` constructor or factory method argument|Types that expose a constructor or `….of(…)`/`….valueOf(…)` factory method taking a `Streamable` as argument. See xref:repositories/query-methods-details.adoc#repositories.collections-and-iterables.streamable-wrapper[Returning Custom Streamable Wrapper Types] for details.
-|Vavr `Seq`, `List`, `Map`, `Set`|Vavr collection types. See xref:repositories/query-methods-details.adoc#repositories.collections-and-iterables.vavr[Support for Vavr Collections] for details.
+|Vavr `Seq`, `List`, `Map`, `Set`|Vavr collection types, also accepted as query method parameters. See xref:repositories/query-methods-details.adoc#repositories.collections-and-iterables.vavr[Support for Vavr Collections] for details.
 |`Future<T>`|A `Future`. Expects a method to be annotated with `@Async` and requires Spring's asynchronous method execution capability to be enabled.
 |`CompletableFuture<T>`|A `CompletableFuture`. Expects a method to be annotated with `@Async` and requires Spring's asynchronous method execution capability to be enabled.
 |`Slice<T>`|A sized chunk of data with an indication of whether there is more data available. Requires a `Pageable` method parameter.


fResult:
Hi, @AzazelSensei,
Thanks so much for taking this and submitting the PR!

I have a couple of suggestions to make the documentation clearer and cover the items mentioned in the #3525 issue:

## 1. Refine phrasing in [`query-methods-details.adoc`:L281-283](https://github.com/AzazelSensei/spring-data-commons/blob/cc3441e76db840239f291831158fea4a2611da08/src/main/antora/modules/ROOT/pages/repositories/query-methods-details.adoc?plain=1#L281-L283)

```diff
  That is, a `java.util.List` is turned into a Vavr `List` or `Seq`, a `java.util.Set` becomes a Vavr `LinkedHashSet` `Set`, and so on.

+ These Vavr types are also accepted as query method *parameters* (as of 2.0). If a repository method parameter uses one of the types above (or `Traversable`), Spring Data automatically unwraps it into its Java-native equivalent before executing the query.
+
+ For example, a derived query method declared as `findAllByIdIn(io.vavr.collection.List<Long> ids)` accepts a Vavr `List` and unwraps it to a `java.util.List` so it binds correctly as an `IN`-clause. A Vavr `Option` parameter is similarly unwrapped to its inner value (or `null`).

  The same types are also accepted as query method *parameters*.
```


## 2. Update table entry in [`query-return-types-reference.adoc`:L30](https://github.com/AzazelSensei/spring-data-commons/blob/cc3441e76db840239f291831158fea4a2611da08/src/main/antora/modules/ROOT/pages/repositories/query-return-types-reference.adoc?plain=1#L30)

```diff
- |Vavr `Seq`, `List`, `Map`, `Set`|Vavr collection types. See xref:repositories/query-methods-details.adoc#repositories.collections-and-iterables.vavr[Support for Vavr Collections] for details.
+ |Vavr `Seq`, `List`, `Map`, `Set`|Vavr collection types, also accepted as query method parameters. See xref:repositories/query-methods-details.adoc#repositories.collections-and-iterables.vavr[Support for Vavr Collections] for details.
```

Let me know what you think.
And thx again for your help! 🙌

AzazelSensei:
Took your wording. The Vavr section now has the 2.0 unwrap, Traversable, and the findAllByIdIn example. Table row mentions parameters too.

mp911de:

Issue #3522, opened 2026-08-07
`Jsr310Converters` use `systemDefault` timezone


jandurovec:
## Problem

The converters for `LocalTime`, `LocalDate` and `LocalDateTime` currently convert the local values to a `Date` (to be stored in the DB) using `systemDefault()` timezone, e.g. https://github.com/spring-projects/spring-data-commons/blob/0ba14c1adc092e910ce20376bc7842f6b53e877c/src/main/java/org/springframework/data/convert/Jsr310Converters.java#L165

This works OK if writing and reading of the value happens in the same JVM or in JVMs running in the same location.

However, for an application distributed across timezones (and with shared DB) it creates an issue, because when the data saved by one instance are read by another instance, they retrieve different information.

### Verification

A sample test case would look something like this

```java
@Test
void preservesLocalTimeWhenReadInDifferentSystemDefaultTimeZone() {

        var defaultTimeZone = TimeZone.getDefault();
        var source = LocalTime.of(12, 34, 56);

        try {
                TimeZone.setDefault(TimeZone.getTimeZone(ZoneId.of("Europe/London")));
                var stored = CONVERSION_SERVICE.convert(source, Date.class);

                TimeZone.setDefault(TimeZone.getTimeZone(ZoneId.of("America/New_York")));
                var read = CONVERSION_SERVICE.convert(stored, LocalTime.class);

                assertThat(read).isEqualTo(source);
        } finally {
                TimeZone.setDefault(defaultTimeZone);
        }
}
```

Currently, this case would fail.

## Proposal

I would suggest using a fixed (UTC) timezone when converting  `LocalTime`, `LocalDate` and `LocalDateTime` to `Date` values, which would bring consistency when being stored and read regardless of the timezone.

The issue is that this is a breaking change, and applications that already have some data in the database (stored in their timezone) will break when upgrading the lib version to the one with the fix, because they will suddenly read different information (unless they run in JVM with UTC timezone)

As a workaround, this can be solved by the clients by adding `@Field(targetType = STRING)` on their entities, but in this case `StringToLocalTimeConverter` should be added to this library, because it is missing (converters from `String` to `LocalDate` and `LocalDateTime` exist)

adityaanikam:
[...]

PR #3535, opened 2026-08-24
Add `String` to `LocalTime` converter

jandurovec:
The current set of JSR310 converters is missing the support for deserialization of `LocalTime` from `String`.

See #3522


mp911de:

PR #3537, opened 2026-08-29
Avoid query rewrite allocations without expressions

### LINKED ISSUE #3536: Avoid allocations when rewriting queries without value expressions
Spring Data Commons 4.2.0-SNAPSHOT constructs a mutable expression map, a rewritten query string, and a second quotation map for every query passed to `ValueExpressionQueryRewriter`, even when the query contains no value expression. Ordinary repository queries commonly take this path.

A no-expression fast path can retain the initial matcher and quotation-map construction, then reuse the original query string and an immutable empty parameter map when the first matcher search finds nothing. This preserves unbalanced-quotation validation and quotation metadata.

JMH on JDK 26, with three forks and the GC profiler, measured the ordinary-query path at 608.007 B/op before and 232.004 B/op after the fast path, a deterministic 61.8% allocation reduction. Independent negative controls were unchanged to measurement precision: quoted expression-like text remained 1,368 B/op and an actual expression remained 9,576 B/op.

A separate 729-query fresh-JVM batch used 30 forks, no warmup, SingleShotTime, and 99.9% confidence intervals. It improved from 7.577 ± 0.845 ms and 1.422 MB to 5.349 ± 0.201 ms and 1.008 MB, reducing both time and allocation by about 29%.

The proposed change includes focused tests for input-string reuse, retained quotation information, unbalanced quotations, real expressions, and expression-like text inside quotes, plus a permanent JMH benchmark.

gregjotau:
fix #3536

### DIFF
--- a/src/jmh/java/org/springframework/data/repository/query/ValueExpressionQueryRewriterBenchmarks.java
+++ b/src/jmh/java/org/springframework/data/repository/query/ValueExpressionQueryRewriterBenchmarks.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2026-present the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


gregjotau:
The team can take over this. I dont' need ownership, I just want stuff done and efficient ;)

